PR #1, opened 2015-03-24
Work in progress, to-date

petemoore:


schema.json
```diff
@@ -0,0 +1,122 @@
...
+    },
+    "maxRunTime": {
+      "type": "number",
+      "title": "Maximum run time in seconds",
+      "description": "Maximum time the task container can run in seconds",
+      "multipleOf": 1.0,
+      "minimum": 1,
+      "maximum": 86400
+    },
+    "graphs": {
```
jonasfj: Uou should not add support for this initially... If at all ever! :)

schema.json
```diff
@@ -0,0 +1,122 @@
...
+      "type": "array",
+      "title": "Paths (in the container) to a json files which are used to extend the task graph",
+      "description": "Contents of file are used to extend the graph (if this task was part of a graph). See http://docs.taskcluster.net/scheduler/api-docs/#extendTaskGraph",
+      "items": {
+        "title": "Individual path to graph extension point.",
+        "type": "string"
+      }
+    },
+    "artifacts": {
```
jonasfj: You don't have to support `artifacts` initially, but it's very very nice to have :)

Either way, if you do I would suggest making it an array of objects instead of a mapping. But as I see it, you can just leave it up to the tasks to upload artifacts for now...

schema.json
```diff
@@ -0,0 +1,122 @@
...
+          "path",
+          "expires"
+        ]
+      }
+    },
+    "features": {
+      "title": "Feature flags",
+      "description": "Used to enable additional functionality.",
+      "type": "object",
+      "properties": {
```
jonasfj: drop most of this.... If not all... at least initally.

model.go
```diff
@@ -0,0 +1,155 @@
...
+	RunId             uint                    `json:"runId"`
+	QueueMessage      QueueMessage            `json:"-"`
+	SignedURLPair     SignedURLPair           `json:"-"`
+	ClaimCallSummary  queue.CallSummary       `json:"-"`
+	reclaimTimer      *time.Timer             `json:"-"`
+	TaskClaimRequest  queue.TaskClaimRequest  `json:"-"`
+	TaskClaimResponse queue.TaskClaimResponse `json:"-"`
+	Definition        queue.TaskDefinition1   `json:"-"`
+	Payload           GenericWorkerPayload    `json:"-"`
+}
```
jonasfj: I would split this struct more... (I haven't read all the code yet).

But I would use structures:

 * `Advice: {taskId, runId, QueueMessage, SignedUrlPair}` as contents of a queue message
 * `TaskClaim: {taskId, runId}` as data from when you have obtained a task claim
 * `TaskRun: {taskId, runId, definition, payload, reclaimTimer, takenUntil}`

Not sure it's perfect, just a suggestion....

signedurlmanager.go
```diff
@@ -0,0 +1,65 @@
...
+		for _, q := range signedURLs.Queues {
+			log.Println("  Delete URL: " + q.SignedDeleteUrl)
+			log.Println("  Poll URL:   " + q.SignedPollUrl)
+		}
+	}
+	// Get signed urls for the first time...
+	updateUrls()
+	// loop forever, serving requests for signed urls, or requests to refresh
+	// signed urls since they are about to expire...
+	for {
```
jonasfj: So you keep a fresh URL around forever... 
If you are working on a task that takes hours... This seems unnecessary...

I think you can do
```go
func (signedURLs *queue.PollTaskUrlsResponse) refresh() {
  if (signedURLs.Expires.Sub(time.Second * time.Duration(premInt)).Before(time.Now())) {
    signedURLs, _ = Queue.PollTaskUrls(os.Getenv("PROVISIONER_ID"), os.Getenv("WORKER_TYPE"))
  }
});
```

Then in `main.go` you just do:
```js
signedURLs, _ = Queue.PollTaskUrls(os.Getenv("PROVISIONER_ID"), os.Getenv("WORKER_TYPE"))
```

And before each use you just ensure that it's up to date by calling: `signedURLs.refresh()`.

**Warning,** I'm a go newbie... So I'm not sure you can just define a method for `queue.PollTaskUrlsResponse` in another module, but I suspect it just works :)

Eitherway, there is no reason to keep valid queue poll urls around when working on a task, and no need to not make that check and request synchronously.

But you probably should set `REFRESH_URLS_PREMATURELY_SECS` to 5 min to allow for slow requests and retries...

main.go
```diff
@@ -0,0 +1,543 @@
...
+		return nil, err
+	}
+	reader := strings.NewReader(string(fullBody))
+	dec := xml.NewDecoder(reader)
+	err = dec.Decode(&queueMessagesList)
+	if err != nil {
+		return nil, err
+	}
+	if len(queueMessagesList.QueueMessages) == 0 {
+		return nil, errors.New("Zero tasks returned in Azure XML QueueMessagesList")
```
jonasfj: This is not really an error... I would just return `nil, nil`... and treat as no message available... it's an expected case, using the error code path can be unpleasant...

main.go
```diff
@@ -0,0 +1,543 @@
...
+
+	// at this point we know there is precisely one QueueMessage (== task)
+	qm := queueMessagesList.QueueMessages[0]
+
+	// To find the task referenced in a message the worker must base64
+	// decode and JSON parse the contents of the <MessageText> tag. This
+	// would return an object on the form: {taskId, runId}.
+	m, err := base64.StdEncoding.DecodeString(qm.MessageText)
+	if err != nil {
+		return nil, err
```
jonasfj: If this fail, you should delete the azure queue message... otherwise another worker will pick it up and it'll keep being thrown around until the task deadline is reached where which is the timetolive for the message.

main.go
```diff
@@ -0,0 +1,543 @@
...
+// translates the Azure response into a Task object
+func (urlPair SignedURLPair) Poll() (*TaskRun, error) {
+	queueMessagesList := new(QueueMessagesList)
+	// To poll an Azure Queue the worker must do a `GET` request to the
+	// `signedPollUrl` from the object, representing the Azure queue. To
+	// receive multiple messages at once the parameter `&numofmessages=N`
+	// may be appended to `signedPollUrl`. The parameter `N` is the
+	// maximum number of messages desired, `N` can be up to 32.
+	// Since we can only process one task at a time, grab only one.
+	resp, err := http.Get(urlPair.SignedPollUrl + "&numofmessages=1")
```
jonasfj: You must implement retries with exponential backoff, I'm sure there is a go library for doing exactly this.

jonasfj: It's not critical here... because it's inside a loop :)

main.go
jonasfj: Hence, move the base64 decode step down to just before you do the JSON decode...

main.go
```diff
@@ -0,0 +1,543 @@
...
+	err = json.Unmarshal(m, &taskRun)
+	if err != nil {
+		return nil, err
+	}
+
+	// Workers should read the value of the `<DequeueCount>` and log messages
+	// that alert the operator if a message has been dequeued a significant
+	// number of times, for example more than 10.
+	if qm.DequeueCount > 10 {
+		log.Printf("WARN: Task with id %v has been dequeued %v times!", taskRun.TaskId, qm.DequeueCount)
```
jonasfj: Do this before you decode the JSON message.
In fact do it as soon as you have constructed the `signedDeleteUrl`.
If the message has been dequeued `15+` times, you should log and delete the message, remember to delete with retries...

main.go
```diff
@@ -0,0 +1,543 @@
...
+
+	// initialise fields of TaskRun not contained in json string m
+	taskRun := TaskRun{
+		QueueMessage:  qm,
+		SignedURLPair: urlPair,
+	}
+	// now populate remaining json fields of TaskRun from json string m
+	err = json.Unmarshal(m, &taskRun)
+	if err != nil {
+		return nil, err
```
jonasfj: If JSON decode fails, you should delete the message with retries...

main.go
```diff
@@ -0,0 +1,543 @@
...
+	// operations fails with a 4xx error. A 400 hundred range error implies
+	// that the task wasn't created, not scheduled or already claimed, in
+	// either case the worker should delete the message as we don't want
+	// another worker to receive message later.
+
+	req, err := http.NewRequest("DELETE", task.SignedURLPair.SignedDeleteUrl, nil)
+	if err != nil {
+		return err
+	}
+	resp, err := http.DefaultClient.Do(req)
```
jonasfj: retries with exponential backoff...

main.go
```diff
@@ -0,0 +1,543 @@
...
+	// repeatedly fails to be deleted, it would result in a lot of unnecessary
+	// calls to the queue and the Azure queue. The worker will likely continue
+	// to work, as the messages eventually disappears when their deadline is
+	// reached. However, the provisioner would over-provision aggressively as
+	// it would be unable to tell the number of pending tasks. And the worker
+	// would spend a lot of time attempting to claim faulty messages. For these
+	// reasons outlined above it's strongly advised that workers logs failures
+	// to delete messages from Azure queues.
+	if !(200 <= resp.StatusCode && resp.StatusCode < 300) {
+		return fmt.Errorf("Status code for http delete request not in 2xx range: %v", resp.StatusCode)
```
jonasfj: You might want to only log `404` errors...

main.go
```diff
@@ -0,0 +1,543 @@
...
+	// Using the taskId and runId from the <MessageText> tag, the worker
+	// must call queue.claimTask().
+	tcrsp, callSummary := Queue.ClaimTask(task.TaskId, fmt.Sprintf("%d", task.RunId), &task.TaskClaimRequest)
+	task.ClaimCallSummary = *callSummary
+
+	// check if an error occurred...
+	if callSummary.Error != nil {
+		// If the queue.claimTask() operation fails with a 4xx error, the
+		// worker must delete the messages from the Azure queue.
+		if callSummary.HttpResponse.StatusCode/100 == 4 {
```
jonasfj: Don't delete if you have a 401 authentication error :)

Generally, if it's not a 404 or 409 error you should probably log the error... but still delete just for good measure :)

main.go
```diff
@@ -0,0 +1,543 @@
...
+	}
+	// When the worker has completed the task successfully it should call
+	// `queue.reportCompleted`.
+	return task.reportCompleted()
+}
+
+func (task *TaskRun) prepEnvVars() {
+	for i, j := range task.Payload.Env {
+		fmt.Printf("Setting env var %v to '%v'\n", i, j)
+		os.Setenv(i, j)
```
jonasfj: This can't be the  way of doing it... I'm sure one can set the environment variables of a subprocess to be different from current process.... Imagine a task overwrite worker config by mistake :)

main.go
```diff
@@ -0,0 +1,543 @@
...
+	cmd.Stderr = os.Stderr
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+	log.Printf("Waiting for command to finish...")
+	// start a go routine to kill task after max run time...
+	go func() {
+		time.Sleep(time.Second * time.Duration(task.Payload.MaxRunTime))
+		task.cancel("max run time (" + strconv.Itoa(task.Payload.MaxRunTime) + "s) exceeded")
```
jonasfj: Probably better named abort... but I guess this is just `cmd.Process.Kill()`.

When you add logging support it might be nice if you after killing writing a message to the log, saying that it was killed...

main.go
jonasfj: See `Cmd.Env` property at http://golang.org/pkg/os/exec/#Cmd

jonasfj:
I'm totally new to go, so let me know what is wrong with my comments :)

model.go
jonasfj: Having read everything else... I see that using one bug struct is convenient, though it forces you to remember which order things are constructed in.

To make it simpler, one could make a struct with sub-structs...  but the flow is fairly sane so I would just go with this... :)

main.go
petemoore: In general I haven't implemented them because:

* the worker tries to grab tasks every 1 second (and exponential backoff would slow this down)
* the task only gets deleted if all the http calls are successful, so in the case where an http call fails, a second later the same worker will have tried to claim it again

In other words, I think the whole process gets retried every second, so an exponential backoff won't really save us much, as all the http requests are pre task-run

The exception I guess is post flight http requests, once the task is in progress. Then there is a strong need for exponential backoff, so that compute time isn't wasted (e.g. after an hour of running a task, we don't want a failed http request to require that job starts from beginning again).

Do you agree with this choice? I could be oversimplifying, let me know what you think.

main.go
petemoore: See above - I wonder if delete with retries is really necessary, if it will in any case retry to delete one second later? That said, maybe some intermediate http calls could be saved by doing with retries upfront. I wonder if maybe I should put the retry support directly in taskcluster-client-go library...

main.go
jonasfj: > Do you agree with this choice? I could be oversimplifying, let me know what you think.
I think the simplification will work...

Personally I would do retries with exponential backoff even if it slows down, because can very well be desired result...
Also maybe polling every 5-10s is better... Or just make it configurable.

Just make sure all other API calls are retried... Hence, you might as well implement a generic method for retrying stuff. You can probably do a retryFunction(callback, TRANSIENT_ERRORS) function where callback is a function that does the request and returns any error... and TRANSIENT_ERRORS is a list of transient error codes...
Mostly because you'll critically need it for the delete message from azure call..

Note, the go taskcluster client library should do the retries already (right), if not we should definitely fix that first.

main.go
jonasfj: You must call `queue.claimTask` before deleting the message... So definitely retry the delete operation.

Basically not retrying a request with exponential backoff an extreme corner case that should only be used in very very rare cases... Even if exponential backoff slows down the request rate that can be important so the server has time to recover.

For something like a service that runs on heroku that has to return API calls within 30s, we like to make retries quick... still exponential, but very quick... I think it's up to 6s sleep or so... While things that run as background workers should generally sleep longer for retries... and definitely retry. It improves robustness and reduces obscurity.
So even in cases where you can argue that it's not critical, you might as well just to keep a good pattern going :)


> I wonder if maybe I should put the retry support directly in taskcluster-client-go library...
If `taskcluster-client-go` doesn't implement retries with exponential backoff that is a bug.
Client retry logic is implemented here: https://github.com/taskcluster/taskcluster-client/blob/master/lib/client.js#L292-L303

With the following configuration:
https://github.com/taskcluster/taskcluster-client/blob/master/lib/client.js#L46-L55

Using `delayFactor:    100,` is the same as AWS, and great for servers that needs to reply within 30s, but for background processes and workers I would recommend 500ms instead... There is no reason a worker shouldn't be patient.

If workers are too eager they could overload the servers and if backoff isn't exponential they would able to keep the servers down. So `delayFactor:    500,` is probably a good idea for a background worker, as it gives delays: 500ms, 1s, 2s, 4s, 8s, 16s (with 5 retries). Such delays should give the servers some air so that they have a chance of getting on top. Note, we typically do 1 requests followed by up to 5 retries by default, and default to `delayFactor:    100,`, but encourage background processes and workers to use `delayFactor:    500,`.

Funny story when pulse was down for 10 min, queue.taskcluster.net experienced 50k failed requests... Probably because of retries. Because normally we don't see that many requests, anyways when that many clients keep retrying it's nice that retries are exponential, or servers might crash as soon as they come online again.

Note, one might even argue that we should implement some random factor of delay, i.e multiply the delay for each retry as follows `delay = delay * (Math.random() * 0.2 + 0.9)´.

----
@pmoore, I realize that this is not important until we have a lot of generic workers running. So I'm mostly emphasising the importance of retry logic, to ensure that we have good practices on this subject.

PR #2, opened 2015-12-04
Prevent "panic: runtime error: hash of unhashable type net.IP"

sheerun:
When loading config such error would occur. Probably because net.IP type
cannot be used as a key in map, and cannot be compared directly. That is why I
switched from array to map and used deep equality of reflect package to check
if default value is set.

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,11 +7,13 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"reflect"
 	"strings"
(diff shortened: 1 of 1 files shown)


petemoore:
Thanks so much!! How great to find this pull request, completely out of the blue! :grinning:

I hadn't noticed this was broken, but it certainly was! I'll merge this, and then add some unit tests. I tested locally, and it works beautifully, thank you! :fireworks:

petemoore:
I've now added tests in commit 6c797e21f76675f8023496fde04c80e7d834e240.

Thanks again!

sheerun:
You're welcome. I'm glad to help :)

PR #5, opened 2016-05-13
user home set to default win convention

grenade:
"(OI)(CI)" = This folder, subfolders, and files. (http://ss64.com/nt/icacls.html)


petemoore:
Sorry, looks like I didn't spot this PR in time. This is now achieved differently by creating a user profile programatically, which also takes care of creating dirs with correct permissions for home folder / profile path.

PR #6, opened 2016-05-25
Adapted to work with upstream changes in taskcluster-client-go

petemoore:
Upstream removed `CallSummary` and made some lint fixes.


jonasfj:
Fro a quick look I say r+, but I don't have a firm grasp on generic worker... it seems like a small change anyways...

PR #7, opened 2016-06-01
Renamed all_commands.log to live_backing.log to fix links sent by mozilla-taskcluster to treeherder

petemoore:
Thanks Greg!

### DIFF
--- a/livelog/livelog.go
+++ b/livelog/livelog.go
@@ -5,9 +5,11 @@ package livelog
 
 import (
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
+	"strconv"
 	"time"
 
 	"github.com/taskcluster/slugid-go/slugid"
@@ -68,6 +70,10 @@ func New(liveLogExecutable, sslCert, sslKey string) (*LiveLog, error) {
 	}
 	l.setRequestURLs()
 	err = l.connectInputStream()
+	// Note we can't wait for GET port to be active before returning
+	// since livelog will only serve from that port once some content is
+	// sent - so no good to execute waitForPortToBeActive(60023) here...
+	// We would need to fix this in livelog codebase not here...
 	return l, err
 }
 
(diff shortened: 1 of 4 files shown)


gregarndt:
I would be happy to review this once you have the changes pushed for handling live logging redirect artifacts.  feel free to assign me once you got something up! :)

livelog/livelog.go
```diff
@@ -97,13 +103,25 @@ func (l *LiveLog) connectInputStream() error {
 	}
 	client := new(http.Client)
 	go func() {
```
gregarndt: Why are we running this in a go routine? Is it so the task starts and continues on while this waits for the live log to start?

livelog/livelog.go
```diff
@@ -97,13 +103,25 @@ func (l *LiveLog) connectInputStream() error {
...
 		// ignore any error and response we get back...
 		client.Do(req)
 	}()
 	return nil
 }
+
+func waitForPortToBeActive(port int) {
+	deadline := time.Now().Add(60 * time.Second)
+	for time.Now().Before(deadline) {
+		conn, err := net.DialTimeout("tcp", "localhost:"+strconv.Itoa(port), 60*time.Second)
```
gregarndt: doesn't DialTimeout block until either there is a connection or the timeout is reached?

petemoore: It is a timeout for the connection handshake. However, if the connection is actively refused, it will exit with an error. Therefore we keep the timeout, in case the connection handshake never completes, but we also have a loop to retry in case the listener is not yet active.

livelog/livelog.go
petemoore: Yes, exactly.

livelog/livelog.go
gregarndt: Ah thanks, i didn't know that.  Did you discover that from documentation or by looking at the code? I think I might have missed something.

gregarndt:
👍

livelog/livelog.go
petemoore: I wrote some code to test the behaviour, and discovered that was how it worked. It started a go routine which slept a few seconds and then started an http server, and in another go routine, it ran the DialTimeout call, which exited immediately. Then I adapted it to look as above, and that worked. I didn't commit the test, just because tests that sleep didn't seem like a good idea, but I guess I could have committed it with short timeouts (like 50ms or so). In the end though, no doubt we'll replace this with TaskCluster worker so maybe not worth the effort of writing extensive tests for this.

PR #8, opened 2016-06-03
Bug 1277568: redirect livelog artifact to underlying log when task command completes

petemoore:
Hey Greg,

This should mirror docker worker behaviour, so that after a live log is no longer available, the livelog artifact is replaced by an alternative redirect artifact which points to the underlying log file.

The only difference is that we have one livelog per command in the Generic Worker, rather than just a single live log per task.

### DIFF
--- a/main.go
+++ b/main.go
@@ -813,15 +813,7 @@ func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
 		return WorkerShutdown(err)
 	}
 
-	// the duration until the expiry is precisely -1 *
-	// the amount of time since the expiry
-	duration := -time.Since(time.Time(task.Definition.Expires))
-	commandLogURL, err := Queue.GetArtifact_SignedURL(
-		task.TaskId,
-		fmt.Sprintf("%v", task.RunId),
-		task.Commands[index].logFile,
-		duration,
-	)
+	commandLogURL := fmt.Sprintf("%v/task/%v/runs/%v/artifacts/%v", Queue.BaseURL, task.TaskId, task.RunId, url.QueryEscape(task.Commands[index].logFile))
 
 	if err != nil {
 		return WorkerShutdown(err)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -812,6 +812,35 @@ func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
 	if err != nil {
 		return WorkerShutdown(err)
 	}
+
+	// the duration until the expiry is precisely -1 *
+	// the amount of time since the expiry
+	duration := -time.Since(time.Time(task.Definition.Expires))
```
gregarndt: Could this be time.Time(task.Definition.Expires).sub(time.Now()) ?

main.go
```diff
@@ -812,6 +812,35 @@ func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
...
+	commandLogURL, err := Queue.GetArtifact_SignedURL(
```
gregarndt: Wouldn't this cause an expiration issue later? You're using a signed url in the redirect artifact.

main.go
petemoore: Would look a bit nicer, agreed - ok I'll do that.

main.go
petemoore: I did this in case a task had private logs. If they are public, it shouldn't expire, but if they are private, the expiry should match the expiry of the artifact (assuming the worker clock is correct). Until now all logs have been public, but I saw it as safer not to assume they always will be. In the case of the log being protected, if you have the scopes to view the live log, then you should have the scopes to read the log (as the livelog artifact name is the log name + ".live").

gregarndt: If you're using a signed URL, that means anyone viewing the task would be able to retrieve the log.

petemoore: If you don't have required scopes to read the livelog artifact, you won't be able to get the signed URL. If you've retrieved the signed URL, then you already had scopes to retrieve the livelog URL, so you should be allowed access to the s3 url.

But agreed if you leave it with the private artifact path, the person would need to log in, so that still works. I hadn't considered that. So I'll not sign the URL, and keep it simple. Thanks for highlighting a much simpler solution! :+1:

gregarndt:
💯

petemoore:
Thanks Greg! :-)

PR #9, opened 2016-06-29
Bug 1279019: add artifact public/logs/worker_type_metadata.json to each task

petemoore:
This is a place for storing information about how the worker type was set up. In the worker type definition, a json object can be set in `secrets.generic-worker.config.workerTypeMetadata` for adding extra data into `public/logs/worker_type_metadata.json`. Note, it comes with some default fields, which can be overridden or added to:
```json
{
  "generic-worker-version": <generic-worker-version>,
  "go-version":             runtime.Version(),
  "go-arch":                runtime.GOARCH,
  "go-os":                  runtime.GOOS
}
```
It is intended that this would be used for storing a link to the machine setup, _for example_, by adding something like this to the worker type definition:
```
{
  "secrets": {
    "generic-worker": {
      "config": {
        "workerTypeMetadata": {
          "machine-setup": "https://raw.githubusercontent.com/taskcluster/generic-worker/2d2ad3000787f2c893299e693ea3f59287127f5c/worker_types/win2012r2/userdata",
          "maintainer": "pmoore@mozilla.com"
        }
      }
    }
  }
}
```

The idea is that whenever the worker type is updated, this metadata can be refreshed, so anybody running a test can see the metadata about how the worker type was set up that ran their task. This is useful especially on Windows, where the toolchains for the tasks are installed directly in the worker type, rather than in e.g. a docker container like we do on linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
 
 script:
   - "go install -v ./..."
-  - "if test $GIMME_OS.$GIMME_ARCH = linux.amd64; then ${GOPATH}/bin/gotestcover -v -coverprofile=coverage.report ./...; go tool cover -func=coverage.report; fi"
+  - "if test $GIMME_OS.$GIMME_ARCH = linux.amd64; then ${GOPATH}/bin/gotestcover -v -coverprofile=coverage.report ./... && go tool cover -func=coverage.report; fi"
 
 after_script:
   - "$HOME/gopath/bin/goveralls -coverprofile=coverage.report -service=travis-ci"
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ go test -v ./...
 
 1. Bump the version number in `main.go` [here](https://github.com/taskcluster/generic-worker/blob/d1e48692122dd3e295defda1e61acc8509ad7e23/main.go#L58).
 2. Commit the change, e.g. `git add main.go; git commit -m "Bumped version number"`.
-3. Tag the repo, e.g. `git tag v2.0.0alpha2`
+3. Tag the repo, e.g. `git tag v2.0.1alpha2`
 4. Push to github taskcluster repo master branch, e.g. `git push; git push --tags`
 5. Wait for binary releases to magically appear [here](https://github.com/taskcluster/generic-worker/releases) (travis will push them if tests pass).
 
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ go test -v ./...
 
 1. Bump the version number in `main.go` [here](https://github.com/taskcluster/generic-worker/blob/d1e48692122dd3e295defda1e61acc8509ad7e23/main.go#L58).
 2. Commit the change, e.g. `git add main.go; git commit -m "Bumped version number"`.
-3. Tag the repo, e.g. `git tag v2.1.0alpha1`
+3. Tag the repo, e.g. `git tag v2.1.0alpha2`
 4. Push to github taskcluster repo master branch, e.g. `git push; git push --tags`
 5. Wait for binary releases to magically appear [here](https://github.com/taskcluster/generic-worker/releases) (travis will push them if tests pass).
 
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ go test -v ./...
 
 1. Bump the version number in `main.go` [here](https://github.com/taskcluster/generic-worker/blob/d1e48692122dd3e295defda1e61acc8509ad7e23/main.go#L58).
 2. Commit the change, e.g. `git add main.go; git commit -m "Bumped version number"`.
-3. Tag the repo, e.g. `git tag v2.1.0alpha5`
+3. Tag the repo, e.g. `git tag v2.1.0`
 4. Push to github taskcluster repo master branch, e.g. `git push; git push --tags`
 5. Wait for binary releases to magically appear [here](https://github.com/taskcluster/generic-worker/releases) (travis will push them if tests pass).
 
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ go test -v ./...
 
 1. Bump the version number in `main.go` [here](https://github.com/taskcluster/generic-worker/blob/d1e48692122dd3e295defda1e61acc8509ad7e23/main.go#L58).
 2. Commit the change, e.g. `git add main.go; git commit -m "Bumped version number"`.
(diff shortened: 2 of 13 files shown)


jonasfj:
aws instance-id, region, availability-zone

jonasfj:
ip address

jonasfj:
when we have a single log file... please add it to that log file instead of a separate artifact... Artifacts aren't as cheap as 100 byte overlay...

main.go
```diff
@@ -258,6 +267,13 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 		UsersDir:                   "C:\\Users",
 		CleanUpTaskDirs:            true,
 		IdleShutdownTimeoutSecs:    0,
+		WorkerTypeMetaData: map[string]interface{}{
```
gregarndt: instead of creating something generic here with string to an empty interface, why not just create a type for this data?

main.go
```diff
@@ -942,6 +970,21 @@ func (task *TaskRun) run() error {
 	return finalError
 }
 
+func (task *TaskRun) preTaskActions() error {
+	jsonBytes, err := json.Marshal(config.WorkerTypeMetaData)
```
gregarndt: isn't there a json.MarshalIndent()?

model.go
```diff
@@ -18,23 +18,24 @@ type ExecCommand interface {
...
-		RefreshUrlsPrematurelySecs int    `json:"refreshURLsPrematurelySecs"`
-		PublicIP                   net.IP `json:"publicIP"`
-		Subdomain                  string `json:"subdomain"`
-		WorkerGroup                string `json:"workerGroup"`
-		WorkerId                   string `json:"workerId"`
-		WorkerType                 string `json:"workerType"`
-		UsersDir                   string `json:"usersDir"`
-		CleanUpTaskDirs            bool   `json:"cleanUpTaskDirs"`
-		IdleShutdownTimeoutSecs    int    `json:"idleShutdownTimeoutSecs"`
+		AccessToken                string                 `json:"accessToken"`
```
gregarndt: Looking at this, it seems just line formatting changes that snuck in?

petemoore: Yes, that happens automatically because `go fmt` lines up the columns, and there is a new entry added with a type that is longer than the previous types were (`map[string]interface{}`).

main.go
petemoore: We allow any json object for the worker type meta data, so using map[string]interface{} guarantees that regardless of what a worker type creator puts in the worker type definition under this new key, it can be stored, merged with the default set of keys, and then marshalled back to json again.

main.go
petemoore: There is indeed! Good spot! :)

petemoore:
@jonasfj I'll add the fields you suggested. I agree it might be cheaper to put it in the log file, and unite the log files into one, so will aim to do that when we move this to taskcluster-worker.

petemoore:
@jonasfj @gregarndt I'm afraid the generic-worker codebase really is not a pretty thing... A lot of it was created while I was learning go, and it is a big mess. In any case, I got things working for this PR, and it isn't all super pretty, but I think it is good enough to unblock the windows rollout. A lot of this will be cleaned up when we move over to taskcluster-worker. It is ready for review again!

Note, there was a minor issue with the `.travis.yml` that meant that test failures would not result in a non-zero exit code. When I spotted (and fixed) that, I found real failures, which were just specific to running the worker on linux - hence why you'll see some extra changes slipped in here.

Thanks guys!

.travis.yml
```diff
@@ -28,7 +28,7 @@ install:
 
 script:
   - "go install -v ./..."
-  - "if test $GIMME_OS.$GIMME_ARCH = linux.amd64; then ${GOPATH}/bin/gotestcover -v -coverprofile=coverage.report ./...; go tool cover -func=coverage.report; fi"
+  - "if test $GIMME_OS.$GIMME_ARCH = linux.amd64; then ${GOPATH}/bin/gotestcover -v -coverprofile=coverage.report ./... && go tool cover -func=coverage.report; fi"
```
petemoore: ^^^^^ booooo - this hid failures

petemoore:
@jonasfj @gregarndt tested (in production!) [here](https://treeherder.mozilla.org/#/jobs?repo=nss&revision=1e3e144b7e884a296bc5c34dc86490589b180237&filter-searchStr=win)

.travis.yml
gregarndt: wow now that's rather subtle

README.md
```diff
@@ -222,7 +222,7 @@ go test -v ./...
 
 1. Bump the version number in `main.go` [here](https://github.com/taskcluster/generic-worker/blob/d1e48692122dd3e295defda1e61acc8509ad7e23/main.go#L58).
 2. Commit the change, e.g. `git add main.go; git commit -m "Bumped version number"`.
-3. Tag the repo, e.g. `git tag v2.0.0`
+3. Tag the repo, e.g. `git tag v2.1.0`
```
gregarndt: if this is just an example, I don't see why we care about bumping this :)

aws.go
```diff
@@ -31,7 +32,7 @@ func queryUserData() (*UserData, error) {
 	return userData, err
 }
 
-func queryMetaData(url, name string) (string, error) {
+func queryMetaData(url string) (string, error) {
 	// http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-data-retrieval
 	// call http://169.254.169.254/latest/meta-data/instance-id with httpbackoff
```
gregarndt: now that it was changed here, but I'm *very* surprised if you ever had to retry this call.

petemoore: True, that would be surprising! But even if only one in a million times it would fail, I think it would make sense to retry (and the code difference is just `http.Get(url)` vs `httpbackoff.Get(url)`.

README.md
petemoore: I do it because I want people to _feel_ like docs are up-to-date. If they read the example and it is about an ancient release, it will feel like the docs have not been updated since that old release. Making it current just gives the user the feeling that this doc is maintained and relevant, IMHO. It's just a minor detail to hopefully improve trust for the user.

aws.go
gregarndt: oh it was just out of curiosity.  We're silly and just try to get it once within the worker and I wonder if there was something going on that I was missing.

PR #10, opened 2016-06-30
Bug 1181524 - resolve with malformed-payload if artifact.expires < task.deadline

petemoore:
Should probably add some tests...

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


petemoore:
I should also add an artifact with an explanation of why the task resolved with malformed-payload...

gregarndt:
Hrm, with docker-worker we just put the error int he task log.  Woudn't the error just be added to that unified log you create at the end or does it now create that in the case the commands don't run at all?

petemoore:
Good point, thanks Greg! Then I can name the artifact `public/logs/live_backing.log`. Then it is consistent with the other logs (e.g. for treeherder-parsing). In this case there won't be a livelog, but I don't think that matters too much. (Note, the GW only creates the log artifacts if the task runs, which it won't for a malformed-payload, so this shouldn't conflict with any logs the GW would otherwise create).

jonasfj:
An error in the task log would be ideal... but as reporting `malformed-payload` is the most important thing... r+

PR #11, opened 2016-07-01
use a lower case username

grenade:
Due to problems in the Firefox build process (for Windows), the upper case "T" in the username highlights numerous bugs in that build code. This patch bypasses those issues by not creating paths that the build process can choke on. Eg: see https://bugzilla.mozilla.org/show_bug.cgi?id=1280326


petemoore:
Thanks Rob! I think we created at just about the same time, so I'll close this one, as the other has a couple more changes too.

PR #12, opened 2016-07-01
Changed Task_<timestamp> dirs to task_<timestamp>

petemoore:
This is to help out with numerous firefox desktop build bugs. Theoretically we shouldn't need to change generic worker for this, but it is looking so tricky to fix on firefox side, that it just a pragmatic compromise.


petemoore:
@walac Note, I updated `plat_darwin.go` for consistency, but that method isn't actually used, as it was just work in progress from a while ago when I experimented with a darwin implementation.

walac:
r+

PR #13, opened 2016-07-07
Bug 1285197 - fix go get warning for taskcluster-client-go

petemoore:
This depends on taskcluster/taskcluster-client-go#15 so I would expect travis to fail for this PR until that has landed. When the other PR lands, you should be able to rerun the travis jobs on this PR.


jhford:
R+

PR #14, opened 2016-07-08
One livelog per task, instead of per command

petemoore:
To be consistent with docker worker.


jonasfj:
gave it a quick look...seems okay...

jonasfj:
r+

PR #15, opened 2016-07-25
Use credentials from queue.ClaimTask response for uploading artifacts

petemoore:
This should mean we no longer need to grant `queue:create-artifact:*` to all worker types that use generic worker.


jhford:
This looks good to me!

PR #16, opened 2016-08-03
Bug 1291249 - gzip log artifacts and set Content-Encoding in PUT to S3

petemoore:
This now gzip compresses artifact log files. All other artifacts remain not gzip compressed. If all artifacts (not just log files) should be gzip compressed, let me know! :-)

Added check to existing test that artifacts have correct encoding.

### DIFF
--- a/main.go
+++ b/main.go
@@ -884,12 +884,7 @@ func (task *TaskRun) run() error {
 	if err != nil {
 		return WorkerShutdown(err)
 	}
-	// using NewWriterLevel(...) instead of NewWriter(...) to be explicit about
-	// compression level, so easy to change later
-	gzipLogWriter, err := gzip.NewWriterLevel(logFileHandle, gzip.DefaultCompression)
-	if err != nil {
-		return WorkerShutdown(err)
-	}
+	gzipLogWriter := gzip.NewWriter(logFileHandle)
 	gzipLogWriter.Name = "live_backing.log"
 	liveLog, err := livelog.New(config.LiveLogExecutable, config.LiveLogCertificate, config.LiveLogKey)
 	defer func(liveLog *livelog.LiveLog) {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -901,7 +909,7 @@ func (task *TaskRun) run() error {
 		task.logWriter = logFileHandle
```
gregarndt: should this just become the gzip log writer now?

main.go
```diff
@@ -883,6 +884,13 @@ func (task *TaskRun) run() error {
 	if err != nil {
 		return WorkerShutdown(err)
 	}
+	// using NewWriterLevel(...) instead of NewWriter(...) to be explicit about
+	// compression level, so easy to change later
+	gzipLogWriter, err := gzip.NewWriterLevel(logFileHandle, gzip.DefaultCompression)
```
gregarndt: So it seems that NewWriterLevel is almost used instead as a documentation piece to let the developer know that the default level can be overridden? I guess if we were actually overriding it via config or something then it makes sense to use this method, but otherwise this is why the simpler (readability wise) method could be used.  Not big deal though.

main.go
petemoore: Ah yes, nice spot! Thanks Greg!!

main.go
petemoore: Yeah ok I'll switch back. Originally I actually had it on `BestCompression` (hard coded) which worked fine, except that `gunzip` on my mac wasn't able to decompress it, so I switched to the default level. Now I'm doing that, I agree with you, it is no doubt simpler to use `NewWriter` instead.

main.go
```diff
@@ -883,6 +884,13 @@ func (task *TaskRun) run() error {
...
+	if err != nil {
+		return WorkerShutdown(err)
+	}
+	gzipLogWriter.Name = "live_backing.log"
```
petemoore: @gregarndt thought I'd double check - are you setting any other header fields in docker worker that I should also be setting? I assume I should set `Name` here like this in case someone extracts to a file (and I guess this is how the default filename gets chosen if they do). Let me know if this isn't right though!

gregarndt: Really we are not setting anything special....the log is just the task output piped to a gzip module for compression, and then piped to a temp file [1]. Then we're just setting three headers when uploading [2].

[1] https://github.com/taskcluster/docker-worker/blob/master/lib/features/bulk_log.js#L33
[2] https://github.com/taskcluster/docker-worker/blob/master/lib/features/bulk_log.js#L53-L55

gregarndt:
👍

PR #18, opened 2016-09-02
Added more output in cases of generic worker installation failure

petemoore:
@grenade Hope this helps with troubleshooting installation failures. Can you give a quick look and let me know your feedback? Thanks!


grenade:
r+

PR #19, opened 2016-09-02
Bug 1287112 - add chain of trust to generic worker

petemoore:
I still have to update/add tests, and add the missing aws data (private IP address, region, etc) to the generated certificate, but almost there, and certainly in a state others can look at it.

I expect to finish that last bits on Monday.

Please don't merge yet. :-)

CC: @escapewindow @walac


artifacts.go
```diff
@@ -99,45 +100,79 @@ func (errArtifact ErrorArtifact) ResponseObject() interface{} {
...
+	io.Copy(gzipLogWriter, rawContent)
+	gzipLogWriter.Close()
+	return tmpFile.Name(), nil
+}
+
+func (artifact S3Artifact) ProcessResponse(resp interface{}) (err error) {
 	response := resp.(*queue.S3ArtifactResponse)
+	rawContentFile := filepath.Join(TaskUser.HomeDir, artifact.Base().CanonicalPath)
+
+	// if using Content-Encoding: gzip then need to gzip content...
```
walac: nit: s/need/needs/

artifacts.go
```diff
@@ -99,45 +100,79 @@ func (errArtifact ErrorArtifact) ResponseObject() interface{} {
...
 			return nil, nil, err
 		}
-		log.Printf("Content-Type in S3 PUT request: %v", artifact.MimeType)
 		httpRequest.Header.Set("Content-Type", artifact.MimeType)
-		if artifact.ContentEncoding != "" {
-			httpRequest.Header.Set("Content-Encoding", artifact.ContentEncoding)
-			log.Printf("Content-Encoding in put request: %v", artifact.ContentEncoding)
+		httpRequest.ContentLength = transferContentLength
+		if artifact.ContentEncoding == "gzip" {
+			httpRequest.Header.Set("Content-Encoding", "gzip")
```
walac: I'd rather prefer preserving the old behavior, checking if `artifact.ContentEncoding != ""` and then setting `"Content-Encoding"` to `artifact.ContentEncoding`. This way, if in the future we add/change support for a new compression algorithm, it is one less line of code to change.

artifacts.go
```diff
@@ -99,45 +100,79 @@ func (errArtifact ErrorArtifact) ResponseObject() interface{} {
...
 		}
-		// request body could be a) binary and b) massive, so don't show it...
-		requestFull, dumpError := httputil.DumpRequestOut(httpRequest, false)
+		requestHeaders, dumpError := httputil.DumpRequestOut(httpRequest, false)
 		if dumpError != nil {
 			log.Println("Could not dump request, never mind...")
 		} else {
 			log.Println("Request")
-			log.Println(string(requestFull))
+			log.Println(string(requestHeaders))
```
walac: Why the change in the variable name?

artifacts.go
```diff
@@ -99,45 +100,79 @@ func (errArtifact ErrorArtifact) ResponseObject() interface{} {
...
 		}
 		putResp, err := httpClient.Do(httpRequest)
 		return putResp, err, nil
 	}
 	putResp, putAttempts, err := httpbackoff.Retry(httpCall)
 	log.Printf("%v put requests issued to %v", putAttempts, response.PutURL)
 	respBody, dumpError := httputil.DumpResponse(putResp, true)
+	putResp.Body.Close()
```
walac: Any specific reason to not use `defer putResp.Body.Close()` just after line 172? Today it doesn't matter, but it helps maintainability for future changes.

chain_of_trust.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (cot *ChainOfTrust) IsEnabled(fl EnabledFeatures) bool {
+	return fl.GenerateCertificate
+}
+
+func (cot *ChainOfTrust) Killed(task *TaskRun) error {
+	logFile := filepath.Join(TaskUser.HomeDir, "public", "logs", "live_backing.log")
+	certifiedLogFile := filepath.Join(TaskUser.HomeDir, "public", "logs", "certified.log")
+	signedCert := filepath.Join(TaskUser.HomeDir, "public", "logs", "certified.log.gpg")
```
walac: nit: I would do:

```go
logPath := filepath.Join(TaskUser.HomeDir, "public", "logs")
logFile := filepath.Join(logPath, "live_backing.log")
certifiedLogFile := filepath.Join(logPath, "certified.log")
signedCert := filepath.Join(logPath, "certified.log.gpg")
```

Easier to change the log path just in case.

chain_of_trust.go
```diff
@@ -0,0 +1,172 @@
...
+		return
+	}
+	defer in.Close()
+	out, err := os.OpenFile(dst, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		return
+	}
+	defer func() {
+		cerr := out.Close()
+		if err == nil {
```
walac: Isn't it `!=`? Anyway, won't it hide an error from `out.Sync`?

chain_of_trust.go
```diff
@@ -0,0 +1,172 @@
...
+	rawContentFile := filepath.Join(TaskUser.HomeDir, artifact.Base().CanonicalPath)
+	rawContent, err := os.Open(rawContentFile)
+	if err != nil {
+		return
+	}
+	defer rawContent.Close()
+	hasher := sha256.New()
+	_, err = io.Copy(hasher, rawContent)
+	if err != nil {
+		panic(err)
```
walac: A file data copy can fail for several reasons, I don't think we want to `panic` in case of an error here.

main.go
```diff
@@ -246,6 +258,13 @@ func main() {
 			fmt.Printf("%#v\n", err)
 			os.Exit(65)
 		}
+	case arguments["new-openpgp-keypair"]:
+		err := generateOpenPGPKeypair(arguments["--file"].(string))
+		if err != nil {
+			fmt.Println("Error generating OpenPGP keypair for worker:")
+			fmt.Printf("%#v\n", err)
+			os.Exit(66)
```
walac: Just out of curiosity, why `66`?

artifacts.go
petemoore: It contains just the http headers, rather than the full http payload (headers + body). I felt the old name did not reflect that very well, so renamed it.

chain_of_trust.go
petemoore: No, this is called after function has completed. If `err` is not `nil`, it will be returned. However, if it is `nil`, then `cerr` will be returned, which may be `nil` or not `nil`. This is a way to be able to return the error from `out.Close()` even though it is in a `defer` block.

chain_of_trust.go
petemoore: I think if we can't write to the filesystem, we should terminate the worker.

Admittedly we probably want better handling for these edge cases, but probably outside of this PR. I think we should probably have a recover in the main function which shuts down the worker, and then we should probably replace a lot of our `if err != nil { return err }` with `panic(err)` so that we only handle errors that we really can recover from.

main.go
petemoore: I started exit codes at 64, and increased with each one I added. See http://tldp.org/LDP/abs/html/exitcodes.html - in particular:

> The author of this document proposes restricting user-defined exit codes to the range 64 - 113 (in addition to 0, for success), to conform with the C/C++ standard.

chain_of_trust.go
petemoore: I agree your solution has the advantage of isolating the path to one location, but I also see it as having the disadvantage of adding indirection to the code (it is harder to read at a glance where the vars point to). Since the lines are located together, it is easy to edit them at once. I used to prefer to remove any possible source of duplication, but it always needs to be balanced with whether it makes the code less readable. Since text searching in source code is so easy, in this code I prefer to be explicit. Also, there is no guarantee that we'd always want them to all be located under `public/logs` so if one of them moved somewhere else, you'd be required to refactor the code again. This way, you don't have to, so I feel it protects somewhat against future changes, rather than being a hindrance. :-)

petemoore:
@walac I've finished with all my changes now. Feel free to take a look. If you don't see anything bad, please feel free to merge. Thanks!

petemoore:
r+ in bugzilla - merging! Thanks @walac.

PR #20, opened 2016-09-06
show worker type at start of task log

grenade:
with more worker types taking jobs (win 7, 10 testers), it's handy to be able to see quickly which worker type is running the task


petemoore:
# r+ #

PR #21, opened 2016-09-09
Bug 1182451 - preloaded writable directory caches / empty writable directory caches / preloaded readonly directory mounts / preloaded readonly files

petemoore:
This is where I'll be doing the work. Not ready for review - but maybe useful for tracking progress. I'll ping when it is ready for review!

CC @jonasfj @gregarndt

### DIFF
--- a/all-unix-style.yml
+++ b/all-unix-style.yml
@@ -67,6 +67,15 @@ properties:
       - type
       - path
       - expires
+  mounts:
+    type: array
+    title: Mounts
+    description: Directories and/or files to be mounted
+    items:
+      type: object
+      oneOf:
+      - "$ref": "#/definitions/fileMount"
+      - "$ref": "#/definitions/directoryMount"
   features:
     title: Feature flags
     description: Feature flags enable additional functionality.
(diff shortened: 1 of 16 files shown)


petemoore:
@grenade FYI - this PR will enable scope-protected caches

jonasfj:
@petemoore, on irc you said you expanded `assume:` scopes I see nothing doing that.
But I honestly, don't think we should do that... If we wanted to do it, we should probably do it at task-creation time on the queue... Instead of doing it here...
IMO; we 're doing really fine without it.

otherwise, from skimming this it seems pretty good... IMO, tar + zip would probably do fine, but if you had fun adding rar that cool :)
For tc-worker I think I want zstd support like .tar.zst, as it's really fast... But tar.bz2 is awesome too :)

petemoore:
| @petemoore, on irc you said you expanded assume: scopes I see nothing doing that.

I slipped it into https://github.com/taskcluster/taskcluster-base-go/blob/82cb0421bd23ae0a7eb6fd56e59b9b41cdfca99c/scopes/scopes.go#L85-L95

| IMO; we 're doing really fine without it.

Ah, I didn't realise we're not doing it in the other workers - I just assumed we were!

| If we wanted to do it, we should probably do it at task-creation time on the queue... Instead of doing it here...

I'm not sure - we have one queue, but thousands of workers - I think handling it in the worker is potentially less of a bottleneck. Also I implemented it in the scopes library (our go equivalent of `taskcluster-lib-scopes`) and hopefully this should catch all uses. Note, it only expands scopes, if the scopes are not satisfied when not expanding scopes, and then only if it finds at least one scope that beings with "assume:". One disadvantage - it isn't so nice that this library call can now go off and make API requests. I at least made ScopeExpander an interface so it can be mocked/replaced but it still isn't ideal, and now can return an error, which it couldn't before...

Another advantage of expanding in the queue is that it should only happen once. One disadvantage is that if the roles change after the task is submitted, the changes would not be reflected in the task. That said, it does make the builds more deterministic, so probably that is a good thing.

The more I think about this, the more I agree with you... Probably a job for the queue. Maybe we should change this in a separate bug, as while the queue doesn't expand scopes, I think overall it is better that the worker does it. As soon as the queue does it, I think we can disable it in the worker (or in the taskcluster-base-go library). Agree?

petemoore:
@jonasfj You know ... if we made it illegal for a required scope to start with `assume:` (which might make @djmitche happy) then we could be sure that we never expanded scopes twice, since the expanded scopes could be changed to not include the `assume:...` scope it expanded from.

djmitche:
| IMO; we 're doing really fine without it.

No, we do [expand scopes in docker-worker](https://bugzilla.mozilla.org/show_bug.cgi?id=1269443).  I think the expansion is important, and I think that doing so in the queue is going to cause some confusion and may increase the size of tasks dramatically.

| Note, it only expands scopes, if the scopes are not satisfied when not expanding scopes, and then only if it finds at least one scope that beings with "assume:".

This feels pretty purpose-specific, duplicating logic that's in the auth service, and particularly around scopes that makes me nervous.

petemoore:
@djmitche 

> > Note, it only expands scopes, if the scopes are not satisfied when not expanding scopes, and then only if it finds at least one scope that beings with "assume:".

> This feels pretty purpose-specific, duplicating logic that's in the auth service, and particularly around scopes that makes me nervous.

It is a huge worker performance gain. If no scopes begin with `assume:` then there is no need to make an expensive network call to the auth service. Instead of taking ~ 1 second to check whether scopes are satisfied, it takes ~1 millisecond. Another gain is load on the auth service itself, especially when we might have thousands of workers running at once. Note, the code to check for `assume:` prefix is quite straightforward. I think it is worth the gain. If nothing else, it will cut down the time to run unit tests from > 1 minute to a few milliseconds. :-)

djmitche:
How many checks are you performing for each task?? 60??

jonasfj:
@petemoore, okay, I realize that expansion is expected.
@djmitche, we have to do scope-subset checking locally after expansion, so attempting that before expansion makes sense. As for whether or not to check for any `assume:` scopes before expanding that is a bit complicated, and unlikely to give much performance. Negative results shouldn't be expected, as scope issues should be uncommon :)

petemoore:
Taking over `r+` from bugzilla.

PR #22, opened 2016-09-19
bug 1302257; retry taskdirs delete a few times; r?pmoore

grenade:
not terribly confident this will work, but hopeful. based on https://blog.abourget.net/en/2016/01/04/my-favorite-golang-retry-function


petemoore:
Closing, as we decided against this approach.

PR #23, opened 2016-09-23
Bug 1305048 - Add node 6.6.0 to win2012r2.

Standard8:
I believe these commands should do it. I followed what is already done for BinScope.


petemoore:
Looks perfect! A quick test seemed to be successful: http://people.mozilla.org/~pmoore/Screen%20Shot%202016-09-23%20at%2016.51.46.png so is probably good. I suspect the `PATH` will be correctly configured for the task users, but if not, the installation seems to be in `C:\Program Files\nodejs` so we can try adding that afterwards, if it didn't work.

Thanks for the patch! 👍

PR #24, opened 2016-09-29
Fixed several issues that came up in testing of caches/mounts

petemoore:
This PR is for stuff that I work on during heavier testing of caches/mounts...


petemoore:
@jonasfj @gregarndt FYI... still need to add some more tests (for purge cache) but otherwise I think we're good.

petemoore:
Note, this was a lot of cleanup, but this is new: https://github.com/taskcluster/generic-worker/blob/96c8529460b9495ddb3f3b8a7e817c6e4c6f7f13/mounts.go#L601-L648

petemoore:
yes! :-)

PR #25, opened 2016-10-05
Bug 1306988 - add feature "OS Groups" 

petemoore:
@grenade This should allow tasks to specify extra OS groups that their task user should be made a member of.


jonasfj:
looks sane to me from a quick skim...

PR #26, opened 2016-10-28
Bug 1313622 - Livelog command output fix

petemoore:
@grenade This seems to fix it. Writing an automated is a little bit tricky, as I need to make sure that I connect to the livelog stream after the livelog starts, but before the livelog gets replaced by a redirect artifact to the backing livelog, and ideally I need to do that in a different go routine (thread) to the running worker so signalling is not so trivial to make sure that happens, and that I can retrieve the livelog secret in time to get the right livelog url.

Anyways, I did a manual test: [run one](https://tools.taskcluster.net/task-inspector/#LrtCzl_BTAq2WpH_23aRXA/0) with the fix applied and [run two](https://tools.taskcluster.net/task-inspector/#RlW3riYWRSiwmAmM5BCjHQ/0) with the fix not applied, and the patch worked as expected. Without the patch, the command output was missing, but with the patch applied, the command output was streamed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -394,7 +394,7 @@ func runWorker() {
 		if r := recover(); r != nil {
 			log.Printf("Shutting down immediately - panic occurred!")
 			log.Println(string(debug.Stack()))
-			log.Printf("Cause: %#v", r)
+			log.Printf("Cause: %v", r)
 			if config.ShutdownMachineOnInternalError {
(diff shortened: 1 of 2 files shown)


petemoore:
Whoops, didn't mean to close this. Anyway, it's on master now.

PR #28, opened 2016-11-04
Use runlib for managing processes in windows

petemoore:
This isn't working yet - but creating the PR for visibility....


petemoore:
I made an alpha release, and am testing it with [this NSS task](https://tools.taskcluster.net/task-inspector/#fJflHoN-TMOUR1gdYC63pg/0). All seems to be ok!

petemoore:
@walac I don't expect a thorough review, as it is a huge rewrite. I think the only way to test this really is in production. Note - previously we weren't running win7/win10 tests - which is where we want to test this, so there is no concern that releasing something non-perfect will have a big impact. The worst that can happen is that we still don't have a working worker, which is no worse than we are now. But the testing so far has been positive, so I'd propose we just roll out, and if there are issues, work through them then, one by one.

PR #30, opened 2016-11-21
bug 1312383 - set scheduled task process priority to above normal

grenade:
not sure it will help, but the current scheduled task runs its process with priority 7 (below normal) while priority 0 (real time) is the highest, 10 (idle) is the lowest.
https://msdn.microsoft.com/en-us/library/windows/desktop/aa383512(v=vs.85).aspx


petemoore:
This looks good. I noticed this changes the line endings from `CRLF` to `LF` in the xml file - but I'm guessing that shouldn't be a problem. Thanks for contributing this!

PR #40, opened 2017-02-09
Bug 1261188 - set user profile path

grenade:
i'm hoping that setting these values at user creation causes the LoadUserProfile system call to create the AppData folder under the task folder rather than under the os default user profile directory when the task directory is in a non-standard location


petemoore:
@grenade I think we decided this isn't needed anymore - do you agree?

petemoore:
Closing, as discussed on IRC.

PR #42, opened 2017-02-23
Use gometalinter

petemoore:
Let's see how this works out...


djmitche:
@petemoore could we use `golangci-lint` instead?

PR #45, opened 2017-04-18
Retrieve tasks using claimWork

gregarndt:
Curious if this is all it takes to use claimWork now.  Tests passed locally, but wanted to see the results from the CI.

Roles will need queue:claim-work:\<provisioner id\>/\<worker type\> and queue:worker-id:test-worker-group/*


README.md
```diff
@@ -250,31 +247,6 @@ Simply run:
 generic-worker --config CONFIG-FILE
 ```
 
-and watch logs for a successful startup. If you can see it is polling the Queue, and the process does not exit, then you can continue. If it reports a problem, follow any instructions it provides. If you are really stuck, join #taskcluster channel on irc.mozilla.org, and ask for help.
```
gregarndt: I only removed this because it seemed much of the logging was specific about Azure, which no longer should happen.  I've been trying to get a config put together to run the worker locally to see what's logged now to put something back in here, but haven't had luck so far.

petemoore:
This looks solid - and I'd forgotten how much code was involved, so will be really nice to strip this all out.

I'd like to investigate the failures before merging, but it looks very promising. :+1:

gregarndt:
I'm not sure what I did to cause the OS X failures, but they've been happening from the start.

The recent Windows failure is odd, it's in TestMounts (meaning it claimed a task, which is what I touched), so I'm going to retrigger it again to see what happens.

petemoore:
Thanks @gregarndt !!! \o/

PR #49, opened 2017-04-19
Bug 1347956 - gzip more artifacts

petemoore:
This now uses a blacklist of file extensions to _not_ gzip compress, everything else gets gzip content-encoding. Based on this [blacklist](https://github.com/evansd/whitenoise/blob/03f6ea846394e01cbfe0c730141b81eb8dd6e88a/whitenoise/compress.py#L21-L29).

CC @edmorley @gregarndt @jonasfj

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -73,8 +73,7 @@ tasks:
           source ~/env_private.sh
           go get -d "github.com/taskcluster/generic-worker"
           cd "${GOPATH}/src/github.com/taskcluster/generic-worker"
-          git remote add '{{ event.head.user.login }}' '{{ event.head.repo.url }}'
-          git fetch --tags '{{ event.head.user.login }}'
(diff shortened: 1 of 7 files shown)


petemoore:
The single push CI failure was a service interruption when `go get` was run against a repository, so was an intermittent issue, rather than a real failure, so merging.

jonasfj:
@petemoore, have you considered brotli compression instead of gzip?
Also could imagine file extensions:
```
7z
s7z
lz
lz4
lzo
lzma
sz
z
xz
apk
cab
jar
rar
```
I just took what I guess was most common from: https://en.wikipedia.org/wiki/List_of_archive_formats

jonasfj:
@petemoore, brotli is faster and better than gzip, and it support as an internet standard...

edmorley:
Brotli is awesome for assets being consumed by browsers (eg Treeherder serves both gzipped and Brotli compressed versions of it's static assets, thanks to WhiteNoise), however these artifacts have consumers such as curl/requests/... which don't yet support Brotli.

PR #51, opened 2017-04-25
Setup correct PWD environment variable for child processes. r=pmoore

walac:
PWD environment variable is generally handled by the shell.
Nevertheless, scripts may use it in strings that references environment
variables.

### DIFF
--- a/plat_all-unix-style.go
+++ b/plat_all-unix-style.go
@@ -99,6 +99,8 @@ func (task *TaskRun) EnvVars() []string {
 	for i, j := range taskEnv {
 		taskEnvArray = append(taskEnvArray, i+"="+j)
 	}
+	// PWD environment variable may be used for spawned scripts
+	taskEnv["PWD"] = taskContext.TaskDir
 	log.Printf("Environment: %#v", taskEnvArray)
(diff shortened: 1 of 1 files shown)


petemoore:
@walac I believe this isn't needed. Can you give an example of where this is needed, or something that broke?

As I see it, if `$PWD` is used in a script, it implies the process be a descendent of a shell process, i.e. the pertinent command in the task payload should being with e.g. `/bin/bash` or `/bin/sh` etc. If a process is spawned that expects `$PWD` to be set, but is not spawned in a shell, there are many other things that will potentially break, so I think we shouldn't simulate this one feature of the shell, since this is just one of many shell features (e.g. no shell builtins will work) - instead the task should be invoked from a shell.

If I'm missing something, let me know - an example would be good of something that is broken without this fix, that isn't fixed by just executing a shell. Thanks!

petemoore:
@walac See https://tools.taskcluster.net/task-inspector/#E2EUHfamQ5CK8W4__5W8IQ/0 as an example - when invoking a shell, PWD is set, as it should be, but not when executing a binary directly, which is also correct.

PR #54, opened 2017-05-25
Port t-w monitor package to g-w.

walac:
Based on taskcluster-worker monitor package, we implement monitoring
support on generic-worker.

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -121,7 +121,8 @@ func (artifact *S3Artifact) CreateTempFileForPUTBody() string {
 	baseName := filepath.Base(rawContentFile)
 	tmpFile, err := ioutil.TempFile("", baseName)
 	if err != nil {
-		log.Fatal(err)
+		incidentID := monitor.ReportError(err, "Could not create temporary file")
+		log.Fatalf("%v, incident ID: %s", err, incidentID)
 	}
 	defer tmpFile.Close()
 	var target io.Writer = tmpFile
--- a/main.go
+++ b/main.go
@@ -47,6 +47,7 @@ var (
 		&ChainOfTrustFeature{},
 		&MountsFeature{},
 	}
+	monitor *Monitor
 
 	version = "10.0.5"
 	usage   = `
@@ -195,6 +196,8 @@ and reports back results to the queue.
                                             [default: 60023]
           numberOfTasksToRun                If zero, run tasks indefinitely. Otherwise, after
                                             this many tasks, exit. [default: 0]
+          project                           The project name used in [sentry](https://sentry.io/welcome/).
+                                            [default: "test-monitor"]
           provisionerId                     The taskcluster provisioner which is taking care
                                             of provisioning environments with generic-worker
                                             running on them. [default: test-provisioner]
@@ -203,6 +206,12 @@ and reports back results to the queue.
                                             when each task starts. If it cannot free enough
                                             disk space, the worker will shut itself down.
                                             [default: 10240]
+          runAfterUserCreation              A string, that if non-empty, will be treated as a
+                                            command to be executed as the newly generated task
+                                            user, each time a task user is created. This is a
+                                            way to provide generic user initialisation logic
+                                            that should apply to all generated users (and thus
+                                            all tasks).
           runTasksAsCurrentUser             If true, users will not be created for tasks, but
                                             the current OS user will be used. Useful if not an
                                             administrator, e.g. when running tests. Should not
@@ -236,12 +245,6 @@ and reports back results to the queue.
                                             the worker type will have more information about how
                                             it was set up (for example what has been installed on
                                             the machine).
-          runAfterUserCreation              A string, that if non-empty, will be treated as a
-                                            command to be executed as the newly generated task
-                                            user, each time a task user is created. This is a
-                                            way to provide generic user initialisation logic
-                                            that should apply to all generated users (and thus
-                                            all tasks).
 
     If an optional config setting is not provided in the json configuration file, the
     default will be taken (defaults documented above).
@@ -324,7 +327,21 @@ func main() {
 			log.Printf("%v\n", err)
 			os.Exit(64)
 		}
-		exitCode := RunWorker()
+		monitor = NewMonitor(
+			config.Project,
+			auth.New(&tcclient.Credentials{
+				ClientID:    config.ClientID,
+				AccessToken: config.AccessToken,
+				Certificate: config.Certificate,
+			}),
+			"debug",
+			nil,
+		)
+
+		exitCode := INTERNAL_ERROR
+		incidentID := monitor.CapturePanic(func() {
+			exitCode = doRunWorker()
+		})
 		log.Printf("Exiting worker with exit code %v", exitCode)
 		switch exitCode {
 		case REBOOT_REQUIRED:
@@ -337,7 +354,7 @@ func main() {
 			}
 		case INTERNAL_ERROR:
 			if config.ShutdownMachineOnInternalError {
-				immediateShutdown("generic-worker internal error")
+				immediateShutdown("generic-worker internal error, incident ID: " + incidentID)
 			}
 		case NONCURRENT_DEPLOYMENT_ID:
 			immediateShutdown("generic-worker deploymentId is not latest")
@@ -386,6 +403,7 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 		LiveLogPUTPort:                 60022,
 		LiveLogGETPort:                 60023,
 		NumberOfTasksToRun:             0,
+		Project:                        "test-monitor",
 		ProvisionerID:                  "test-provisioner",
 		RefreshUrlsPrematurelySecs:     310,
 		RequiredDiskSpaceMegabytes:     10240,
@@ -453,6 +471,7 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 		{value: c.WorkerGroup, name: "workerGroup", disallowed: ""},
 		{value: c.WorkerID, name: "workerId", disallowed: ""},
 		{value: c.WorkerType, name: "workerType", disallowed: ""},
+		{value: c.Project, name: "project", disallowed: ""},
 	}
 
 	for _, f := range fields {
@@ -492,6 +511,11 @@ func RunWorker() (exitCode ExitCode) {
 		}
 	}()
 
+	exitCode = doRunWorker()
+	return
+}
+
+func doRunWorker() (exitCode ExitCode) {
 	log.Printf("Detected %s platform", runtime.GOOS)
 	// number of tasks resolved since worker first ran
 	// stored in a json file, since we may reboot between tasks etc
@@ -509,6 +533,7 @@ func RunWorker() (exitCode ExitCode) {
 		log.Printf("OH NO!!!\n\n%#v", err)
 		panic(err)
 	}
+
 	// Queue is the object we will use for accessing queue api
 	Queue = queue.New(
 		&tcclient.Credentials{
--- a/monitor.go
+++ b/monitor.go
@@ -1,0 +1,248 @@
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	raven "github.com/getsentry/raven-go"
+	"github.com/pborman/uuid"
+	"github.com/taskcluster/statsum"
+	tcauth "github.com/taskcluster/taskcluster-client-go/auth"
+)
+
+// NewMonitor creates a new monitor
+func NewMonitor(project string, auth *tcauth.Auth, logLevel string, tags map[string]string) (m *Monitor) {
+	// Create statsumConfigurer
+	statsumConfigurer := func(project string) (statsum.Config, error) {
+		res, err := auth.StatsumToken(project)
+		if err != nil {
+			return statsum.Config{}, err
+		}
+		return statsum.Config{
+			Project: res.Project,
+			BaseURL: res.BaseURL,
+			Token:   res.Token,
+			Expires: time.Time(res.Expires),
+		}, nil
+	}
+
+	// Create logger and parse logLevel
+	logger := logrus.New()
+	switch strings.ToLower(logLevel) {
+	case logrus.DebugLevel.String():
+		logger.Level = logrus.DebugLevel
+	case logrus.InfoLevel.String():
+		logger.Level = logrus.InfoLevel
+	case logrus.WarnLevel.String():
+		logger.Level = logrus.WarnLevel
+	case logrus.ErrorLevel.String():
+		logger.Level = logrus.ErrorLevel
+	case logrus.FatalLevel.String():
+		logger.Level = logrus.FatalLevel
+	case logrus.PanicLevel.String():
+		logger.Level = logrus.PanicLevel
+	default:
+		panic(fmt.Sprintf("Unsupported log-level: %s", logLevel))
+	}
+
+	// Convert tags to logrus.Fields
+	fields := make(logrus.Fields, len(tags))
+	for k, v := range tags {
+		fields[k] = v
+	}
+
+	// Declare monitor so we can reference it in OnError
+	m = &Monitor{
+		Statsum: statsum.New(project, statsumConfigurer, statsum.Options{
+			OnError: func(err error) { m.ReportWarning(err) },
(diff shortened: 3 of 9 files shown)


jonasfj:
Note. I was thinking you would just import from tc-worker... But this is decent too... Since this doesn't use govendor...

mounts.go
```diff
@@ -335,6 +335,11 @@ func (taskMount *TaskMount) Start() *CommandExecutionError {
 		// If the problem is internal (e.g. can't mount a writable cache) then
 		// this is handled by a panic.
 		if err != nil {
+			if monitor != nil {
+				incidentID := monitor.ReportError(err, "Error mounting directory")
+				return Failure(fmt.Errorf("%v, incident ID: %s", err, incidentID))
+			}
+
```
petemoore: I think we should only report errors for exceptions, not failures (this would be a task failure, and can be e.g. if an invalid url is specified for the preloaded content of the mount). There should be other places where we explicitly panic in the code.

main.go
```diff
@@ -388,6 +404,9 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 		TasksDir:                       defaultTasksDir(),
 		WorkerGroup:                    "test-worker-group",
 		WorkerTypeMetadata:             map[string]interface{}{},
+		Project:                        "",
+		LogLevel:                       "debug",
+		Tags:                           "",
```
petemoore: Should tags be a string array, rather than a string?

Also, all these config variables need to be defined in the help text at the top of `main.go`. Note config is defined in two sections - mandatory and optional parameters - make sure to add to the correct section.

main.go
```diff
@@ -388,6 +404,9 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 		TasksDir:                       defaultTasksDir(),
 		WorkerGroup:                    "test-worker-group",
 		WorkerTypeMetadata:             map[string]interface{}{},
+		Project:                        "",
```
petemoore: Since empty project is not allowed, maybe default can be something like `"test-project"`.

main.go
```diff
@@ -315,7 +317,21 @@ func main() {
...
+				Certificate: config.Certificate,
+			}),
+			config.LogLevel,
+			nil,
+		)
+
+		exitCode := INTERNAL_ERROR
+		incidentID := monitor.CapturePanic(func() {
+			exitCode = RunWorker()
+		})
```
petemoore: I think I prefer capturing the panic inside `RunWorker` as before, since all of the tests use `RunWorker` as their entry point - thus, we'd lose the panic handling in tests if we do it here. The main method is very slim and well tested, so I do not think it is possible for a panic to occur there.

main.go
```diff
@@ -464,14 +485,6 @@ func RunWorker() (exitCode ExitCode) {
...
 	}
 
-	defer func() {
-		if r := recover(); r != nil {
-			log.Print(string(debug.Stack()))
-			log.Print(" *********** PANIC occurred! *********** ")
-			log.Printf("%v", r)
-			exitCode = INTERNAL_ERROR
-		}
-	}()
```
petemoore: As mentioned above - all the tests call `RunWorker()` so let's keep the panic handling inside of it.

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	raven "github.com/getsentry/raven-go"
+	"github.com/pborman/uuid"
+	"github.com/taskcluster/statsum"
+	tcauth "github.com/taskcluster/taskcluster-client-go/auth"
+)
+
+// NewMonitor creates a new monitor
```
petemoore: I think this comment serves no purpose. Could you elaborate on what a monitor is, whether it can be called concurrently, how it should be used, etc?

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+		panic(fmt.Sprintf("Unsupported log-level: %s", logLevel))
+	}
+
+	// Convert tags to logrus.Fields
+	fields := make(logrus.Fields, len(tags))
+	for k, v := range tags {
+		fields[k] = v
+	}
+
+	// Declare monitor so we can reference it in OnError
```
petemoore: Note, you can achieve this too by declaring it as a return variable, i.e.

```
func NewMonitor(project string, auth *tcauth.Auth, logLevel string, tags map[string]string) (m *Monitor) {
```

That also simplifies the return statement to just `return`.

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+
+type sentry struct {
+	client     *raven.Client
+	m          sync.Mutex
+	project    string
+	expiration time.Time
+	auth       *tcauth.Auth
+}
+
+func (s *sentry) Client() (*raven.Client, error) {
```
petemoore: Here `sentry` is not exported, so `Client()` is not available outside the package, so maybe this should be `client()`? Note, the generic-worker is horribly inconsistent everywhere, but I just wanted to call it out as it seemed strange having an unexported type with an exported method (or is it satisfying an interface?)

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+			return nil, err
+		}
+		// Set expiration, so we remember to refresh
+		s.expiration = time.Time(res.Expires)
+	}
+
+	return s.client, nil
+}
+
+type Monitor struct {
```
petemoore: A comment here would be useful, also for the members (e.g. what is the prefix, why are tags map[string]string here, but in the json it is just a string?)

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+		m.Error("Failed to send error: ", err)
+		return
+	}
+
+	// Send packet
+	_, done := client.Capture(packet, tags)
+	<-done
+}
+
+func (m *Monitor) WithTags(tags map[string]string) *Monitor {
```
petemoore: Doc comment would be useful here.

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (m *Monitor) ReportWarning(err error, message ...interface{}) string {
+	incidentID := uuid.NewRandom()
+	m.Entry.WithField("incidentId", incidentID.String()).WithError(err).Warn(message...)
+	m.submitError(err, fmt.Sprint(message...), raven.WARNING, incidentID, 1)
+	return incidentID.String()
+}
+
+func (m *Monitor) submitError(err error, message string, level raven.Severity, incidentID uuid.UUID, skipFrames int) {
```
petemoore: Doc comment would be useful here.

monitor.go
```diff
@@ -0,0 +1,223 @@
...
+
+type Monitor struct {
+	*statsum.Statsum
+	*logrus.Entry
+	*sentry
+	tags   map[string]string
+	prefix string
+}
+
+func (m *Monitor) Time(name string, fn func()) {
```
petemoore: Doc comment would be useful here (I know it is simple, but I've no idea what the statsum Time function does).

monitor_test.go
```diff
@@ -0,0 +1,50 @@
...
+	require.True(t, incidentID == "")
+
+	incidentID = m.CapturePanic(func() {
+		callingSomethingBad()
+	})
+	require.True(t, incidentID != "")
+}
+
+func badThingHappens() {
+	panic("Oh, this is bad")
```
petemoore: love it!

petemoore:
Also please request a second review when you've made the changes - and this time I promise to be much quicker reviewing - I'm really sorry the last one took so long to get around to (it arrived just in the middle of a complete overhaul/redesign of the worker). Thanks!

petemoore:
P.P.S. Thanks for not importing taskcluster-worker as I think in this case, pulling it out was the cleaner and better solution. I think [Rob Pike would agree](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s). :-)

main.go
walac: I don't remember exactly why, I remember I moved `CapturePani`c here because doing it inside `RunWorker` would make things more complicated.

main.go
walac: This is to make sure whoever is using g-w remembers to set up a proper project name.

monitor.go
walac: I adapted the code from tc-worker, so this could make sense there but not here.

mounts.go
walac: iirc, I copied the behavior from tc-worker, which reports fatal errors.

main.go
petemoore: What about for dev environments, where they don't want to hit production sentry etc?

mounts.go
petemoore: This isn't a fatal error - it is just a task failure.

main.go
petemoore: If a monitor instance isn't created during tests, is this handled gracefully?

The RunWorker method is designed not to panic, so I don't like it being able to panic now, which is a change in behaviour that could affect test results.

main.go
petemoore: I prefer either:

a) not hitting statsum / sentry if a value isn't defined (which is probably what we want for dev environments)
b) using a default like `test-monitor` such that we have a defined bucket for assigning scopes (e.g. the role for running tests includes the scope to use `test-monitor` in the same way it grants the use of `test-provisioner` etc).

main.go
walac: > If a monitor instance isn't created during tests, is this handled gracefully?
Do you mean monitoring calls?

>  The RunWorker method is designed not to panic, so I don't like it being able to panic now, which is a change in behavior that could affect test results.
Hrm, if a test expects `RunWorker` to panic, shouldn't it capture it? `Panic` is something orthogonal and unexpected by nature, so caller either shouldn't expect any function to panic or expect any function to...

main.go
walac: It is actually not currently used, removed.

monitor.go
walac: Added to the `Monitor` type

artifacts.go
```diff
@@ -121,7 +121,8 @@ func (artifact *S3Artifact) CreateTempFileForPUTBody() string {
 	baseName := filepath.Base(rawContentFile)
 	tmpFile, err := ioutil.TempFile("", baseName)
 	if err != nil {
-		log.Fatal(err)
+		incidentID := monitor.ReportError(err, "Could not create temporary file")
+		log.Fatalf("%v, incident ID: %s", err, incidentID)
```
petemoore: I've only just realised this myself, but I have a few places in the code where I am calling e.g. `log.Fatal` which should really just panic, so that the worker shutdown etc is handled gracefully (since we catch the panics and handle them). This PR is about something else, so I won't ask you to fix all my broken code, but I'll do that myself once this lands. In other words, where there are currently `log.Fatal[f]` function calls that execute _in the main go routine_, I'll change them to `panic` statements instead.

main.go
```diff
@@ -456,14 +476,6 @@ func loadConfig(filename string, queryUserData bool) (*Config, error) {
 }
 
 func RunWorker() (exitCode ExitCode) {
-	log.Printf("Detected %s platform", runtime.GOOS)
```
petemoore: Why has this been removed?

monitor.go
```diff
@@ -0,0 +1,254 @@
...
+// A Monitor is responsible for collecting logs, stats and error messages.
+//
+// A monitor is a context aware object for monitoring. That is to say that a
+// Monitor is used to record metrics, write logs and report errors. When doing
+// so the Monitor object adds meta-data to the metrics, logs and errors. The
+// meta-data added is context dependent tags and prefix. These help identify
+// where a log message, metric or error originates from.
+//
+// By encapsulating the context meta-data inside the Monitor object, an
+// implementor gets a Monitor rarely needs to add tags or prefix. For example
```
petemoore: This first sentence doesn't make sense.

monitor.go
```diff
@@ -0,0 +1,254 @@
...
+// a monitor will always be prefixed by plugin name before being passed to a
+// plugin, hence, it is easy trace any log message, metric or error report to
+// the plugin that it was created in.
```
petemoore: Can you reword "plugin" as "feature"? ("plugin" is the name we use in taskcluster-worker, but in generic-worker we called the equivalent thing a "feature")

petemoore:
> Note. I was thinking you would just import from tc-worker... But this is decent too... Since this doesn't use govendor...

@jonasfj I think it would be a shame to pull in the entire taskcluster-worker repo just for this. See also [this](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s).

petemoore:
@walac Also, can you clarify the reason that we have the panic handling in the `main` method rather than just keeping it in the `RunWorker` function (and why to add the doRunWorker function)? Thanks!

petemoore:
@walac Also looks like there are merge conflicts that need resolving... Thanks!

walac:
> @walac Also, can you clarify the reason that we have the panic handling in the `main` method rather than just keeping it in the `RunWorker` function (and why to add the doRunWorker function)? 
When running tests, we capture panic through `defer`/`recover`, when running the worker, we need to capture through monitoring. This preserves the previous behavior of `RunWorker`. `doRunWorker` is just a helper to make the code simpler.

main.go
petemoore: Tests don't expect RunWorker to panic, since RunWorker doesn't panic. However, they check the exit code that RunWorker returns, and if a panic occurs, this will be captured in the exitCode. They do check this. So really it is about the RunWorker method being designed to return a code based on what happened (idle timeout, worker requests a reboot, worker crashed, job was successful, etc) - and tests check that. Does that make sense?

petemoore:
The monitor initialisation and panic capture should occur in `RunWorker`, since currently if you call RunWorker (e.g. from a test) we will have a nil pointer exception for any call to a monitor's methods. We don't need doRunWorker - we just need to move the monitor initialisation logic into RunWorker. The global config will always provide the values to use, and in RunWorker, you will always need the values set, so there is no reason to initialise the monitor outside of the RunWorker function, and no reason to create the doRunWorker function.

Additionally, no logging calls have been adapted to use the new Monitor, so none of the existing logging will go through the monitor. If we are not using the logger, let's get rid of it. If we intend to use it, please replace the old logger calls - otherwise we have two different loggers in one codebase.

In the PR I see only a single call on the Monitor object, which is to report a crash to sentry. The other features are implemented, but then not used. I don't see an advantage to port all this code from taskcluster-worker, and then not use it. Since we already have a code block handling exceptions and crashes (panics), it would make more sense to directly report to sentry here, and not introduce all the new abstractions that were designed for a different worker.

In taskcluster-worker there is a concept of having a component hierarchy, which can be used to identify a component when logging information. In generic-worker, there is no such component hierarchy, so the code around this is probably unneeded (since it is not called).

My concern with this PR is it brings over unnecessary complexity from taskcluster-worker, that isn't suited to generic-worker, it provides methods which are never called, and introduces more dependencies. Since it only adds reporting crashes to sentry, maybe it would be best just to add that in a couple of lines of code instead in the current code block that currently logs the panic to papertrail and to the task log.

I don't see an advantage of changing the logging mechanism. The code for reporting to statsum is something that could be useful, but I don't see a good reason for it to be embedded in the monitor object - as far as I see, uploading crash reports is a separate concern to recording activity metrics. The two things use separate systems, there is no code commonality, I think having a Monitor object that handles both these unrelated things makes the design more brittle and introduces false dependencies. Since these are separate concerns, I'd propose we handle them via different interfaces, which can be maintained independently.

In other words, what I'm really asking:

1) let's leave logging as it is, as it does what we want already
2) let's configure a global sentry client, and then use it to report some real (useful) data (i.e. not just create it)
3) let's adapt the existing top-level panic handling routine to also send crash reports to sentry (in addition to what it is already doing)

jonasfj:
@petemoore,
If you don't want the complexity monitor brings with it... just use the
statsum client for golang and sentry client directly.
Neither one is very hard to use... and monitor demonstrates how to do it...

--
Regards Jonas Finnemann Jensen.

2017-06-09 6:26 GMT-07:00 Pete Moore <notifications@github.com>:

main.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/taskcluster/taskcluster-base-go/scopes"
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 	"github.com/taskcluster/taskcluster-client-go/auth"
+	tcauth "github.com/taskcluster/taskcluster-client-go/auth"
```
petemoore: we might not need this

petemoore:
Let's go with this, and delete unused code in a future PR. Thanks!

PR #56, opened 2017-06-12
public/logs/chainOfTrust.json.asc => public/chainOfTrust.json.asc

petemoore:
@Callek @escapewindow This should correct the chain of trust artifact name.


petemoore:
Note, I expect the windows 7 taskcluster CI job is about to fail due to an unrelated problem (open issue) but everything else should pass.

PR #59, opened 2017-07-31
Bug 1385870 - Do not gzip content encode artifacts with .xz extension

lissyx:


artifacts_test.go
```diff
@@ -462,6 +462,12 @@ func TestUpload(t *testing.T) {
 			contentEncoding: "", // jpg files are blacklisted against gzip compression
 			expires:         payload.Artifacts[0].Expires,
 		},
+		"SampleArtifacts/b/c/d.tar.xz": {
+			extracts:        []string{},
+			contentType:     "application/x-xz",
+			contentEncoding: "", // xz files are blacklisted against gzip compression
+			expires:         payload.Artifacts[0].Expires,
+		},
```
petemoore: Unfortunately, this test won't work, since the file hasn't been added to the SampleArtifacts directory. Can you revert the change to `artifacts_test.go` and just keep the change to `artifacts.go`?
Thanks!

lissyx: Should be okay now @petemoore :)

petemoore:
Thanks!

PR #61, opened 2017-08-04
Bug 1383024 - Superseding

petemoore:
This initial implementation is simpler than the docker-worker implementation.  I have chosen an initial simpler implementation for the following reasons:

1. it is simpler, and therefore less likely to break or contain bugs, which is important as I will be on PTO for two weeks
2. I also believe it is a better implementation

Differences to docker-worker implementation

Docker worker claims all tasks it finds listed in the superseded service response body. It keeps claims on those tasks while the primary task (the task it has claimed which is not superseded by any other tasks it has claimed) is running. It then resolves all tasks upon completion of the primary task, attaching artifacts to identify which task run superseded a given task run, and which tasks were superseded by a given task run.

The problems I found with this approach are:

1. There is a bug, that if the primary task is aborted (e.g. due to max runtime being exceeded) or the claim is expired (worker crash, loss of connectivity with queue, etc), the non-primary tasks will be "released" by resolving them as worker shutdown, but future runs of those tasks will not employ superseding, since this relies on run 0 of the task being claimed, which has already been resolved as worker-shutdown

2. For users watching their tasks, they will see their task is claimed, and it will appear to have "hung" with no logs being displayed. It will stay in state "running" until the primary task has completed (which could be for example, an hour) without any information about what is going on. This task has been superseded, so will never be run, and so the user may not realise this until an hour later, at which point, if it is a critical task, they will need to run it again (if the superseding task is not sufficient for their purposes). This is a very bad user experience, when the worker knows immediately, that the task is going to be superseded.

3. This is a complex implementation, there are many workflows that have to be considered: task abortion, cancelation, crashes, deadlines expiring of some but not all tasks during task claims. With all these workflows to consider, the chance of bugs is high (for example, I discovered bug described in comment 1, and some others that have been fixed, but there could be several more).

4. "Running" task counts are inaccurate, since it will appear that more tasks are running than actually are. This could affect, among other things, provisioning logic that now or in the future might consider how many available slots there are for executing tasks, and what scaling ratio to use for scaling up worker pools.

5. It is not entirely efficient to claim all of the tasks at once. Consider, that tasks A, B, C are claimed, where C is the primary task. As task C is running, a new task D can come in, which supersedes C. If we had only checked the supersedes URL at the point task C was claimed using claimWork call of the queue, task D may have already been listed, and thus task D only would be run, rather than tasks C and D (i.e. if we can delay execution to when a task is returned by claimWork, rather than explicitly claiming with claimTask).

6. In the case that other bugs were fixed, each time a primary task is aborted/cancelled/claim-expired, a run is "used up" with worker-shutdown exception. If 6 aborts happen for a line of 100 superseded tasks, even with the bug described in comment 1 being fixed, all of these tasks would now be expired/worker-shutdown on all task runs, and we would lose the information that they were actually superseded at all, and have no idea which task superseded them.

7. The docker-worker implementation relies on both queue.claimWork and queue.claimTask, which prevents us from deprecating claimTask.

8. Migrating superseding from the worker into the queue, with this current design, is non-trivial. It means we need a redesign, if we want to move superseding out of the worker and into the queue (which is the better place for it to be, as then it is written once centrally, and applies to all workers globally).

In light of all the above points, it made sense to apply a considerably simpler algorithm, that solved all 8 of these issues(!) The simpler algorithm looks like this:

1. Claim work. Check the supersedingUrl. If the task is not the very first entry in the list, mark it as superseded by the first entry in the list.

 The only thing I see we lose here, is that the task that supersedes other tasks, does not know which tasks it superseded. However, I don't believe this is useful information. If a task is scheduled, and it is executed, the author of that task should not be concerned about what other tasks might have been scheduled that were then superceded, as it bears no consequence on this task. However, the tasks that were superseded still track which task superseded them, which I think is the only path of interest to follow. The author of a task, whose task is superseded, is interested in the task that superseded it, since it holds the result of the task that is relevant to him/her.

Therefore I see only advantages in this approach, and consider it much easier to maintain/reason about, due to its much greater simplicity. It provides quicker feedback, burns down the queue more quickly, is more reliable generally across a broader number of use cases, and doesn't miss anything that the docker-worker implementation offers.

### DIFF
--- a/supersede.go
+++ b/supersede.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"path/filepath"
 
-	"github.com/taskcluster/generic-worker/gwconfig"
+	"github.com/taskcluster/generic-worker/fileutil"
 	"github.com/taskcluster/httpbackoff"
 	"github.com/taskcluster/taskcluster-base-go/scopes"
 )
@@ -62,18 +62,24 @@ func (l *SupersedeTask) Start() *CommandExecutionError {
 	}
 	resp, _, err := httpbackoff.Get(supersederURL)
 	if err != nil {
-		return MalformedPayloadError(err)
+		// if problem with superseder service, let's run all tasks, and not resolve them all as exception
+		l.task.Logf("WARNING: problem accessing supersederUrl: %v", err)
+		l.task.Log("Not able to see if this task has been superseded!")
+		return nil
 	}
 	decoder := json.NewDecoder(resp.Body)
 	var supersedes SupersedesServiceResponse
 	err = decoder.Decode(&supersedes)
 	if err != nil {
-		return MalformedPayloadError(err)
+		// if problem with superseder service, let's run all tasks, and not resolve them all as exception
+		l.task.Logf("WARNING: not able to interpret response from supersederUrl %v as json list of task IDs: %v", supersederURL, err)
+		l.task.Logf("Not able to see if this task has been superseded!")
+		return nil
 	}
 	taskIDs := supersedes.TaskIDs
 	if l.task.TaskID != taskIDs[0] {
 		supersededByFile := filepath.Join(taskContext.TaskDir, supersededByPath)
-		err = gwconfig.WriteToFileAsJSON(
+		err = fileutil.WriteToFileAsJSON(
 			map[string]string{
 				"taskId": taskIDs[0],
 			},
--- a/supersede.go
+++ b/supersede.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	"github.com/taskcluster/generic-worker/gwconfig"
+	"github.com/taskcluster/httpbackoff"
+	"github.com/taskcluster/taskcluster-base-go/scopes"
+)
+
+var (
+	supersededByPath = filepath.Join("generic-worker", "superseded-by.json")
+	supersededByName = "public/superseded-by.json"
+)
+
+type SupersedeFeature struct {
+}
+
+type SupersedesServiceResponse struct {
+	TaskIDs []string `json:"supersedes"`
+}
+
+func (feature *SupersedeFeature) Name() string {
+	return "Supersede"
+}
+
+func (feature *SupersedeFeature) Initialise() error {
+	return nil
+}
+
+func (feature *SupersedeFeature) PersistState() error {
+	return nil
+}
+
+// Supersede is always enabled
+func (feature *SupersedeFeature) IsEnabled(fl EnabledFeatures) bool {
+	return true
+}
+
+type SupersedeTask struct {
+	task *TaskRun
+}
+
+func (feature *SupersedeFeature) NewTaskFeature(task *TaskRun) TaskFeature {
+	task.featureArtifacts[supersededByName] = "superseded feature"
+	return &SupersedeTask{
+		task: task,
+	}
+}
+
+func (l *SupersedeTask) RequiredScopes() scopes.Required {
+	// let's not require any scopes, as I see no reason to control access to this feature
+	return scopes.Required{}
+}
+
+func (l *SupersedeTask) Start() *CommandExecutionError {
+	supersederURL := l.task.Payload.SupersederURL
+	if supersederURL == "" {
+		return nil
+	}
+	resp, _, err := httpbackoff.Get(supersederURL)
+	if err != nil {
+		return MalformedPayloadError(err)
+	}
+	decoder := json.NewDecoder(resp.Body)
+	var supersedes SupersedesServiceResponse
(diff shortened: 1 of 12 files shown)


petemoore:
I'd like to reiterate, of course, that this is just the initial implementation before I go on PTO.

If for any reason, we are not all as happy with the new approach as I am, we can change it. This is initially just to cover the next two weeks.

djmitche:
Summarizing irc discussion, the existing design was for a reason: the superseding task is executed immediately, meaning we're taking advantage of knowledge of the future to do smarter things *now*, rather than to avoid work now.  Imagine a set of queued tasks
```
1 2 A 4 5 6 B 7 8 9 C
```
with C superseding B superseding A.  A worker claims 1, another worker claims 2, and by rights A should start next, and presumably someone is waiting for that result. With the existing design, we observe that executing task C will give us useful information about A and B, and thus it's advantageous to execute C right now.  The result is that task A gets useful results in a normal timeframe (between 2 and 4), and tasks B and C get useful results ahead of time while consuming no resources.

With the design in this PR, results for A and B would be delayed until after task 9, when task C executes.  Furthermore, if a task D is enqueued before C comes to the head of the queue, then A, B, and C will all be further delayed -- a cycle which might continue indefinitely.

Given a constrained pool with priorities, where low-priority (try) jobs don't coalesce, I think this design would limit us to executing approximately one of each type of high-priority task at a time, with all other workers executing low-priority tasks -- in other words, inverting priority and putting try over integration and release trees.

We can try merging this and see what happens on OS X, checking first that the tasks have supersederUrl set, but we should make sure we have the necessary visibility into its behavior and are prepared to back out the change.

petemoore:
Thanks Dustin. Indeed, that highlights a serious flaw in the design, thanks for taking the time to write up the issue. It may (hopefully) be "better than nothing" for the next two weeks, but if not, it can be backed out. I am writing tests at the moment, assuming those all go well, I'll push them into this PR branch.

petemoore:
@djmitche feel free to pull in help from other reviewers if you like

There is some basic test coverage, so I'm pretty confident it is ok in terms of what it intends to do. As discussed, this will need some attention to change the implementation when I'm back from PTO (or anyone can take a look while I'm away too if they are feeling brave).

petemoore:
Note, there is a problem with the win10 worker type which is currently being updated via OCC, so if you see taskcluster CI failures, that is probably the reason. The most important thing is that the travis tests pass. The taskcluster CI is a little broken at the moment.

petemoore:
Note, since I'm not going to be around for the next two weeks, it would be better only to request changes if there is some critical issue (e.g. not for nits, linting, etc).
I bumped go from 1.7.5 to 1.8.3 since I make user of the *http.Server Shutdown method in the new test, which only got added in 1.8. Since 1.8 is only needed for tests, it could still be built with 1.7.5 if required.

petemoore:
Lastly, feel free to play with https://github.com/petemoore/test-supersede/

supersede.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	resp, _, err := httpbackoff.Get(supersederURL)
+	if err != nil {
+		return MalformedPayloadError(err)
+	}
+	decoder := json.NewDecoder(resp.Body)
+	var supersedes SupersedesServiceResponse
+	err = decoder.Decode(&supersedes)
+	if err != nil {
+		return MalformedPayloadError(err)
```
djmitche: This is a confusing status to return in either case (HTTP error or bad JSON), as the payload is not, in fact, malformed.  Instead, just don't supersede -- that's [what docker-worker does](https://github.com/taskcluster/docker-worker/blob/08876fcfa89d600dc2c824b28c0fe72021cf3594/src/lib/task_listener.js#L425).  In general, superseding is designed to fail open, so that an issue with the superseder URL doesn't cause a treeclosure.

supersede.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	taskIDs := supersedes.TaskIDs
+	if l.task.TaskID != taskIDs[0] {
+		supersededByFile := filepath.Join(taskContext.TaskDir, supersededByPath)
+		err = gwconfig.WriteToFileAsJSON(
```
djmitche: Why is a config function used for this?

supersede.go
petemoore: I went with this, as it suggested that the url for the superseding service is not a valid one. That said, if the superseding service craps out, we probably don't want all tasks to resolve as exception - so I'll change this. Nice spot!

supersede.go
petemoore: It just happened to be a utility method that was in the gwconfig package. I could create a separate fileutil package, I guess. Is it ok to leave that for now, as non-critical, and we change that later? Nice spot.

djmitche: yeah, this was just a question, not necessary to change

PR #62, opened 2017-08-23
Bug 1499055 - Initial work on docker engine

petemoore:
Thought I'd see how hard it is to implement a docker engine in generic-worker. Note, we'll need to do the same for taskcluster-worker, so this isn't lost investment, since we need to work out how to do this from go anyway.


.taskcluster.yml
```diff
@@ -407,75 +407,43 @@ tasks:
...
 
   ##########################################################
-  ################### Linux amd64 Build ####################
+  ##################### docker tests #######################
   ##########################################################
 
-  - provisionerId: "{{ taskcluster.docker.provisionerId }}"
-    workerType: "{{ taskcluster.docker.workerType }}"
+  - provisionerId: aws-provisioner-v1
+    workerType: github-worker
```
petemoore: This worker type executes its tasks inside a docker container, but we need to test generic-worker with docker engine, which will require launching docker. We therefore need to run the CI tasks as a native process on the host, rather than inside a docker container, so we'll need a generic-worker host running native engine on linux.

.taskcluster.yml
```diff
@@ -407,75 +407,43 @@ tasks:
...
-          GORACE=history_size=7 CGO_ENABLED=1 go test -timeout 30m -ldflags "-X github.com/taskcluster/generic-worker.revision=$(git rev-parse HEAD)" -v -race ./...
-          "${GOPATH}/bin/ineffassign" .
-      artifacts:
-        public/build/generic-worker-linux-amd64:
-          path: "/go/bin/generic-worker"
-          expires: "{{ '2 weeks' | $fromNow }}"
-          type: file
-      cache:
-        generic-worker-checkout: /go/src
+        - apt-get update && apt-get install -y git golang && mkdir -p "/home/worker/go/src/github.com/taskcluster" && cd "/home/worker/go/src/github.com/taskcluster" && git clone "{{ event.head.repo.url }}" && cd generic-worker && git checkout "{{ event.head.sha }}" && ./docker-test.sh
```
petemoore: It was easier to read when the commands were separated by new lines.

artifacts_docker_test.go
```diff
@@ -0,0 +1,11 @@
+// +build docker
+
+package main
+
+import "github.com/taskcluster/slugid-go/slugid"
+
+var (
+	// all tests can share taskGroupId so we can view all test tasks in same
+	// graph later for troubleshooting
+	taskGroupID = slugid.Nice()
```
petemoore: This is already a feature of the generic-worker tests, why is this logic duplicated?

docker-test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -vex
+
+repo_path=$GOPATH/src/github.com/taskcluster
+
+cd $repo_path
+go get ./...
```
petemoore: why `go get everything under taskcluster`?

docker-test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash -vex
```
petemoore: Why a separate script for this? Why not an option of the existing test script?

dockerworker/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+					d.TaskLogger.Printf("Exiting reclaim loop: %v", err)
+					return
+				}
+				takenUntil = reclaim.TakenUntil
+			case <-d.Context.Done():
+				return
+			}
+		}
+	}()
+}
```
petemoore: Why is this logic in the docker package? Reclaim logic already exists in the worker, and doesn't need to be specialised for the docker engine.

dockerworker/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+
+// DownloadArtifact downloads an artifact using exponential backoff algorithm
+func (d *DockerWorker) DownloadArtifact(taskID, runID, name string, out io.WriteSeeker) (ret error) {
```
petemoore: Why is this logic here too? It also exists already, and is unrelated to docker engine changes.

dockerworker/artifact.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if ret == nil {
+		ret = backoffError
+	}
+
+	return
+}
+
+// UploadArtifact uploads a new artifact to the task
+func (d *DockerWorker) UploadArtifact(taskID, runID, name string, in io.ReadSeeker) (ret error) {
```
petemoore: Same here, I'm unclear why this core, existing functionality has been duplicated here. It doesn't relate to the docker engine, and also exists already.

Can we rename `DockerWorker` to `DockerEngine`? I think that would make it clearer that this is generic-worker with a docker engine, not docker-worker, the other product.

dockerworker/dockerimage_test.go
```diff
@@ -0,0 +1,62 @@
...
+package dockerworker
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	imageTaskID   = "Xx0aPfyOTU2o_0FZnr_AJg"
```
petemoore: What is this taskID, can you provide a comment?

dockerworker/tarstream.go
```diff
@@ -0,0 +1,67 @@
...
+type tarRewriter func(header *tar.Header, r io.Reader) (*tar.Header, io.Reader, error)
+
+// rewriteTarStream rewrites a tar-stream read from r and written to w.
+//
+// For each entry in the tar-stream the rewriter function is called, if it
+// returns an error then the process is aborted, if it returns a nil header the
+// entry is skipped, otherwise the returned header and body is rewritten to the
+// output tar-stream. Notice that rewriter can return the arguments given
+// in-order to let entries pass-through.
+func rewriteTarStream(r io.Reader, w io.Writer, rewriter tarRewriter) error {
```
petemoore: what do we use this for?

dockerworker/renameimage.go
```diff
@@ -0,0 +1,155 @@
...
+// Type to read from manifest.json
+type manifestEntry struct {
+	Config   string   `json:"Config,omitempty"`
+	RepoTags []string `json:"RepoTags"`
+	Layers   []string `json:"Layers"`
+}
+
+// rewriteManifest will rewrite the 'manifest.json' file from a docker image
+// tar-stream such that the image has a single tag named <imageName>:latest
+func rewriteManifest(imageName string, data []byte) ([]byte, error) {
```
petemoore: can you explain what the rewrite functions are used for?

dockerworker/tarstream.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// a tarRewriter is a function that given a tar.Header and io.Reader returns a
+// tar.Header and io.Reader to be written in place of the entry given.
+//
+// If a nil header is returned  the entry is skipped, if no transformation is
+// desired the header and reader given can be returned with any changes.
+type tarRewriter func(header *tar.Header, r io.Reader) (*tar.Header, io.Reader, error)
```
petemoore: This sounds like a general purpose function not specific to the docker engine - can it go in its own package?

Can you provide a bit of context for why this is useful?

.taskcluster.yml
walac: I am giving docker-in-docker a try

artifacts_docker_test.go
walac: `artifacts_test.go` in docker builds

docker-test.sh
walac: It is a bug, thanks for pointing this out

dockerworker/artifact.go
walac: The duplicated functions are used only to test the package (you could argue that they shouldn't be exported then, but it causes no harm either), as they simpler versions and more generic version of the functionality provided by the main package. The integrated code keeps using the generic-worker provided functions.

dockerworker/artifact.go
walac: It is only used by docker to download artifact images and for tests.

dockerworker/artifact.go
walac: DockerEngine is already used by taskcluster-worker, using the name DockerWorker reflects the intention of being fully compatible with docker-worker.

dockerworker/tarstream.go
walac: This is used to rewrite manifests in the docker image. This function was stolen from taskcluster-worker.

dockerworker/renameimage.go
walac: It is used to rewriter the manifest.json inside the downloaded docker image

dockerworker/tarstream.go
walac: When we download a tar docker image, we need to rewrite the manifest.json file before calling `docker load`, and then tar it again.

.taskcluster.yml
walac: I have changed it, I think this comment no longer holds

artifacts_docker_test.go
walac: I fixed it by moving taskGroupID to its own module.

docker-test.sh
walac: It is now inside the docker image used for tests.

dockerworker/dockerimage_test.go
walac: I added a comment explaining what it is.

.taskcluster.yml
```diff
@@ -407,75 +407,45 @@ tasks:
...
 
-  - provisionerId: "{{ taskcluster.docker.provisionerId }}"
-    workerType: "{{ taskcluster.docker.workerType }}"
+  - provisionerId: aws-provisioner-v1
+    workerType: worker-ci-test
     metadata:
-      name: "Build/test 64 bit generic-worker on Debian GNU/Linux 8 (jessie)"
-      description: "This builds the 64 bit linux version of generic-worker"
+      name: "docker-worker tests"
+      description: "This run docker-worker tests"
```
petemoore: runs

aws_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !docker
+
```
petemoore: Do the aws tests not run with the docker engine?

dockerworker/artifact.go
petemoore: The test framework already provides mechanisms to submit tasks, upload artifacts, claim tasks etc, so there is no reason to duplicate this functionality. Please have a look at existing tests to see how this is done. These functions are independent of the engine being used, and the existing code does not need any alteration to be used.

dockerworker/artifact.go
petemoore: OK, please remove it, it isn't needed.

dockerworker/artifact.go
petemoore: In any case, please remove this code, for the reasons stated above.

dockerworker/artifact_test.go
```diff
@@ -0,0 +1,36 @@
...
+	require.Equal(t, content, buff.String())
+
+	// Test Download specific runID
+	buff = filebuffer.New([]byte{})
+	require.NoError(t, d.DownloadArtifact(taskID, "0", "public/test", buff))
+
+	_, err = buff.Seek(0, io.SeekStart)
+	require.NoError(t, err)
+	require.Equal(t, content, buff.String())
+}
```
petemoore: As far as I can see, the docker engine doesn't support artifacts yet. There is no way to mount artifacts, or extract them. Yet this test tests uploading and downloading artifacts, using the methods you wrote, which we established weren't needed, since generic-worker already provides methods to upload/download artifacts.

dockerworker/dockerimage.go
```diff
@@ -0,0 +1,148 @@
+// +build docker
```
petemoore: The management of docker images should really be a task feature, just like the other features we have (os groups, livelog, taskcluster proxy, mounting files/folders/caches, superseding, chain of trust, interactive task, run as administrator, ...). Each feature is called at task start, task stop, and has a chance to do custom initialisation when the worker starts up. This also means task setup steps happen before the max run timer starts, and failures in initialisation/termination can be propagated back to calling code so that appropriate cleanup happens, task resolution is handled correctly, etc.

This is really no different, it is just a mandatory feature if docker engine is specified.

dockerworker/dockerimage_test.go
petemoore: We already have such tasks, and their task definitions are stored in /testdata/tasks - please update that directory and add a note here, if this is needed.

dockerworker/dockerworker.go
```diff
@@ -0,0 +1,70 @@
...
+		TaskLogger:    NewTaskLogger(liveLogWriter),
+		Context:       ctx,
+		Queue:         queue,
+		Client:        cli,
+	}
+}
+
+// ValidatePayload validates the docker worker task payload
+func ValidatePayload(payload json.RawMessage) (result *gojsonschema.Result, err error) {
+	const schema = "https://raw.githubusercontent.com/taskcluster/docker-worker/master/schemas/v1/payload.json"
```
petemoore: We cannot download from this github url for every task that runs in production. The worker should not depend on github.com at runtime, and the load would be enormous. Also it is pointing to a moving target, using the master branch.

dockerworker/tarstream.go
petemoore: I should have been clearer. Why do we need to rewrite manifests? What are we using them for, and why might they need changing?

dockerworker/util_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/taskcluster/slugid-go/slugid"
+	tcclient "github.com/taskcluster/taskcluster-client-go"
+	"github.com/taskcluster/taskcluster-client-go/tcqueue"
+)
+
+func createDummyArtifact(d *DockerWorker, name, content string) (taskID string, err error) {
```
petemoore: unneeded

images/docker-test/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+tags=docker
+
+go get $go_package
+
+cd $GOPATH/src/$go_package
+
+git checkout $repo_sha
+
+go get ./...
+go get -tags=docker ./...
```
petemoore: -tags=$tags

petemoore: and you need `-t`

intermittent_task_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !docker
```
petemoore: You've disabled docker for lots of tests where the engine is not related to the test. This is one example.

process/process_docker.go
```diff
@@ -0,0 +1,146 @@
+// +build docker,!windows
```
petemoore: why not just `+build docker`?

dockerworker/artifact.go
walac: IIUC, the test frame aims black box tests, these utility functions are for white box tests. Since these are used only for tests, I don't think it is a big concern such a small duplication in name of code organization, as long as it doesn't cause regressions. We can always refactor it later.

dockerworker/artifact.go
walac: `DownloadArtifact` is used to download docker images. If you are concerned because it is part of the package public interface, I can make it an internal function.

dockerworker/artifact.go
walac: I have moved this to `util_test.go` and made it a private function to make it clear it is a white box tests helper.

dockerworker/dockerimage_test.go
walac: This is not the same as the `taskID` in the main tests, this is a known task containing the various artifact images formats for tests.

aws_test.go
walac: All tests that use the generic-worker payload won't run in docker builds since it uses the docker-worker payload.

process/process_docker.go
walac: Fixed, thanks.

intermittent_task_test.go
walac: All tests that use the generic-worker payload will fail under docker builds. Since they run in other builds, I disabled them.

images/docker-test/entrypoint.sh
walac: Thanks

dockerworker/util_test.go
walac: This is used for white box tests

dockerworker/dockerworker.go
walac: Oh, I put this as temporary and forgot it, thanks for pointing this out.

petemoore:
> Depends on #128

#130 also fixes the build failures - so a rebase or merge from `master` would unblock this (#130 has already been merged to master).

aws_test.go
petemoore: The plan we had was to add support to generic-worker for running tasks inside a docker container (bugs [1502335](https://bugzil.la/1502335),  [1502337](https://bugzil.la/1502337),  [1502339](https://bugzil.la/1502339),  [1502348](https://bugzil.la/1502348),  [1502351](https://bugzil.la/1502351),  [1502352](https://bugzil.la/1502352)), and _then_ providing a compatibility layer so that docker-worker style payloads would be supported (bug [1502354](https://bugzil.la/1502354)). It sounds like at the moment with these changes, generic-worker style payloads won't work when running a task in docker, which would mean it we don't have a trivial path to align on payload format in future. It also makes the changes more complicated, because all code that touched the payload has to be adapted to work with a different payload format. The idea behind the compatibility layer was that we have an isolated function that converts from one payload format to the other, and all existing code continues to work with the existing data structures, tasks can be interpreted in either format, and we have a simple path for migrating to a standard format.

petemoore:
In the end this was superseded by #175.

PR #63, opened 2017-08-30
Bug 1394557 - retry artifact uploads with HTTP 400 status code response

petemoore:
I decided to retry _all_ HTTP 400 errors, rather than just trying to catch connection inactivity ones, since a production 400 error should be a relatively rare event (generally implying a bug in the code).


petemoore:
Note, I switched to declaring return variables to make the code a little more readable. That makes the patch more noisy, but the only change is returning a temporary (=> intermittent) error if a 400 response is received, to ensure retry kicks in.

Default interpretation of status codes (e.g. 4xx errors are considered permanent, non-retryable errors) only occurs in httpbackoff library if the callback function doesn't return any errors - so just returning a temporary error from the callback is enough to trigger a retry, and override the default status code treatment.

PR #65, opened 2017-09-06
Bug 1397373 - Point to superseding docs

djmitche:


djmitche:
I fixed the `windows.yml` bit.  If you can re-run the generation, that'd be great.

PR #66, opened 2017-09-25
Bug 1402152 - use creds returned from claimWork to reclaim and resolve tasks

petemoore:
Locking shouldn't be required in this case, as generic-worker creates task.Queue on task initialisation in the main go routine, before other go routines are spawned that access it. Therefore there should be no races.

Thanks Jonas!


petemoore:
> Please check that you're also using these credentials for uploading artifacts...

Good point! I checked, and [it is using task credentials](https://github.com/taskcluster/generic-worker/blob/9c8ddef5c11f27441a675a86334634060381c822/artifacts.go#L429-L434). :-)

PR #70, opened 2017-12-08
Bug 1423215 - Make contentType an optional property of file artifact in task payload

petemoore:
This should allow people to explicitly set `Content-Type` of artifact resources if they are not happy with the default choice of the system.

### DIFF
--- a/all-unix-style.yml
+++ b/all-unix-style.yml
@@ -80,6 +80,19 @@ properties:
           description: |-
             Date when artifact should expire must be in the future, no earlier than task deadline, but
             no later than task expiry. If not set, defaults to task expiry.
+        contentType:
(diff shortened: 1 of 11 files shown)


petemoore:
Hey @djmitche feel free to look over this. There are tests, so I'm pretty confident it is ok.

CC @edmorley

petemoore:
> Looks good! I think Ed wants content-encoding, too.

We could also add it (and only allow gzip content encoding). I think for the moment, generic-worker does a reasonable job of deciding if it is worth gzip encoding or not - I guess though that could be useful if you explicitly wanted to disable gzip content encoding. For now I'll land this, and we can have a think about whether we want to add that or not (and also see how it is affected by working with the new artifact api).

PR #73, opened 2018-02-15
Bug 1358545 - check file permissions of chain of trust key and validity of content on startup

petemoore:
1) Don't start worker if the chain of trust key cannot be read or has invalid content.
2) Don't start the worker if the chain of trust key can be read by a task user.


petemoore:
Thanks John!

PR #78, opened 2018-03-09
Bug 1180187 - listen for spot termination notice and abort task on discovery

petemoore:
I think this should do it. Not written any tests yet - will need to mock the termination endpoint to do that, I guess.

### DIFF
--- a/aws.go
+++ b/aws.go
@@ -6,11 +6,11 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io"
 	"io/ioutil"
 	"log"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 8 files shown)


aws.go
```diff
@@ -276,3 +277,24 @@ func deploymentIDUpdated() bool {
...
+		for _ = range ticker.C {
+			resp, err := http.Get("http://169.254.169.254/latest/meta-data/spot/termination-time")
+			// intermittent errors calling this endpoint should be ignored, but can be logged
+			if err != nil {
+				log.Printf("WARNING: error when calling AWS EC2 spot termination endpoint: %v", err)
+				continue
+			}
+			defer resp.Body.Close()
+			if resp.StatusCode == 200 {
+				abort()
```
jonasfj: `break`

jonasfj:
@petemoore, if build failures we had a few days ago was caused by this... then this is NOT a complete fix.
If your worker is terminated by EC2, then your VM will get a shutdown command, similar to pushing the button on your computer. On linux that causes systemd to send `SIGTERM` to all processes, waiting and then sending `SIGKILL`. EC2 will wait a bit after pushing the power button, and then kill your VM by pulling the power plug.

I assume that pushing the power button causes windows to tell all processes to terminate. Thus, your task processes will exit non-zero, and if you're not careful `generic-worker` can report this before respecting the termination signal and shutting down. I'm pretty sure we have this bug in docker-worker.

petemoore:
> @petemoore, if build failures we had a few days ago was caused by this... then this is NOT a complete fix.

@jonasfj No the issues we had recently were before this fix - the fix in this PR should be complete.

PR #82, opened 2018-04-10
Log command result as info not error

petemoore:
See https://mozilla.logbot.info/taskcluster/20180410#c14588279-c14588443
for more details

CC: @Archaeopteryx


petemoore:
Thanks John!

PR #87, opened 2018-05-03
Bug 1458873 - Fix task abortion

petemoore:
See [bug 1458873](https://bugzilla.mozilla.org/show_bug.cgi?id=1458873) for context.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -57,7 +57,7 @@ func TestAbortAfterMaxRunTime(t *testing.T) {
 	}
 	duration := endTime.Sub(startTime).Seconds()
 	if duration < 5 {
-		t.Fatalf("Task %v should have taken at least 3 seconds, but took %v seconds", taskID, duration)
+		t.Fatalf("Task %v should have taken at least 5 seconds, but took %v seconds", taskID, duration)
 	}
 	if duration > 20 {
 		t.Fatalf("Task %v should have taken no more than 20 seconds, but took %v seconds", taskID, duration)
--- a/main_test.go
+++ b/main_test.go
@@ -26,7 +26,11 @@ func TestFailureResolvesAsFailure(t *testing.T) {
 func TestAbortAfterMaxRunTime(t *testing.T) {
 	defer setup(t, "TestAbortAfterMaxRunTime")()
 	payload := GenericWorkerPayload{
-		Command:    sleep(27),
+		Command: append(
+			sleep(27),
+			// also make sure subsequent commands after abort don't run
+			helloGoodbye()...,
+		),
 		MaxRunTime: 5,
 	}
 	td := testTask(t)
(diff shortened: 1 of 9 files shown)


process/process_windows.go
```diff
@@ -119,19 +120,24 @@ func (c *Command) Execute() (r *Result) {
 		r.SystemError = err
 		return
 	}
-	err = c.Wait()
+	// See https://bugzilla.mozilla.org/show_bug.cgi?id=1458873
+	// If we call c.Wait() here, we'll end up waiting for subprocesses
+	// that aren't killed. By calling c.Process.Wait() we only wait for
+	// the parent process to terminate, and not for the stdout/stderr
+	// handles to get closed, which only happens when all subprocesses
+	// have terminated.
```
jhford: Say we're creating a parent process 10, which creates its own child processes 11 and 12.  Does this mean that when we're saying it's OK that we're only killing 10 and possible leaving 11 and 12 alive?

Is there a way that we could see if there are surviving subprocesses and at least log which subprocesses are still living?  Will these be killed when the per-task user is deleted?

main_test.go
```diff
@@ -26,7 +26,11 @@ func TestFailureResolvesAsFailure(t *testing.T) {
 func TestAbortAfterMaxRunTime(t *testing.T) {
 	defer setup(t, "TestAbortAfterMaxRunTime")()
 	payload := GenericWorkerPayload{
-		Command:    sleep(27),
+		Command: append(
```
jhford: nice!

process/process_windows.go
petemoore: > Say we're creating a parent process 10, which creates its own child processes 11 and 12. Does this mean that when we're saying it's OK that we're only killing 10 and possible leaving 11 and 12 alive?

Exactly, for Windows at least. This is not beautiful, but is implemented like this because on Windows 7 nested job groups are not supported, and many gecko processes associate subprocesses with their own job group, so we can't place the processes in our own job group in order to capture the topology. From Windows 8 onwards, nested job groups would allow us to do this, but before Windows 8, it is always possible that subprocesses could be created that we can't track, and therefore on Windows 7 there is no way I could find to guarantee that we kill all subprocesses.

An alternative might have been to iterate through the process tree and find all processes whose stdout/stderr/stdin handles match those of the parent process, and explicitly kill them. This would have the advantage that we've captured all processes which were causing the `Cmd.Wait()` to not return, but like I say, could still leave stray/zombie processes around.

On Windows we necessarily have to reboot between tasks (this will never be optional, since we need a winlogon session to be invoked against the interactive desktop) and therefore we are sure that after the task is resolved, the machine will be rebooted, so any stray processes will get killed. So nothing is really lost. Also for an aborted task, we only really care about the logging up until the time of the abortion. After a reboot, previous task user directories will be cleaned up by the worker, if they could not be cleaned up before the reboot due to file handles remaining open.

> Is there a way that we could see if there are surviving subprocesses and at least log which subprocesses are still living?  Will these be killed when the per-task user is deleted?

Indeed they get killed anyway, due to the reboot. So this PR allows the task resolution/reboot workflow to continue after the parent process has been terminated, and then use the reboot to clear up the rest.

jhford: cool, that addresses my concerns!  It's a shame that windows 7 is such a hard OS to work with sometimes.

PR #88, opened 2018-05-04
Bug 1459203 - Run generic-worker as native service, not via NSSM

petemoore:
Started work on this - still need to sort out logging.


petemoore:
I suspect this has somewhat bit-rotted, and was probably also not implemented correctly or completely, so it probably can be mostly ignored.

PR #89, opened 2018-05-07
Bug 1459376 - log actions taken by mounts feature, and validate SHA256 of content

petemoore:
See [bug 1459376](https://bugzil.la/1459376) for context.

### DIFF
--- a/mounts.go
+++ b/mounts.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"sort"
@@ -70,9 +71,8 @@ type Cache struct {
 	Owner CacheMap `json:"-"`
 	// The key used in the CacheMap
 	Key string `json:"key"`
-	// the size of the file in bytes (cached for performance, as it is
-	// immutable file)
-	// Size int64
+	// SHA256 of content, if a file (not used for directories)
+	SHA256 string `json:"sha256"`
 }
 
 // Rating determines how valuable the file cache is compared to other file
@@ -185,8 +185,8 @@ type TaskMount struct {
 // Represents an individual Mount listed in task payload - there
 // can be several mounts per task
 type MountEntry interface {
-	Mount() error
-	Unmount() error
+	Mount(task *TaskRun) error
+	Unmount(task *TaskRun) error
 	FSContent() (FSContent, error)
 	RequiredScopes() []string
 }
@@ -201,11 +201,15 @@ type FSContent interface {
 	RequiredScopes() []string
 	// Download the content, and return the absolute location of the file. No
 	// archive extraction is performed.
-	Download() (string, error)
+	Download(task *TaskRun) (file string, sha256 string, err error)
 	// UniqueKey returns a string which represents the content, such that if
 	// two FSContent types return the same key, it can be assumed they
 	// represent the same content.
 	UniqueKey() string
+	// SHA256 of the content.
+	RequiredSHA256() string
+	// String representation of where the content comes from
+	String() string
 }
 
 // No scopes required
@@ -322,12 +326,12 @@ func (taskMount *TaskMount) Start() *CommandExecutionError {
 	// services has an outage. Although 1) could not be part of an obscure
 	// attack strategy (although releases shouldn't use caches).
 	if err != nil {
-		taskMount.task.Warn("Could not reach purgecache service to see if caches need purging!")
-		taskMount.task.Warn(err.Error())
+		taskMount.task.Warn("[mounts] Could not reach purgecache service to see if caches need purging:")
+		taskMount.task.Warn("[mounts] " + err.Error())
 	}
 	// loop through all mounts described in payload
 	for _, mount := range taskMount.mounts {
-		err = mount.Mount()
+		err = mount.Mount(taskMount.task)
 		// An error is returned if it is a task problem, such as an invalid url
 		// to download content, or a downloaded archive cannot be extracted.
 		// If the problem is internal (e.g. can't mount a writable cache) then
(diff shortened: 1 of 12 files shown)


petemoore:
@indygreg FYI

indygreg:
You rock for doing this. Thank you so much! Track me down at the SF all hands and I'll buy you a drink, taco, burrito, etc.

petemoore:
> You rock for doing this. Thank you so much! Track me down at the SF all hands and I'll buy you a drink, taco, burrito, etc.

Haha, no worries! I'm happy you raised the bug. 👍

petemoore:
@indygreg Have a look over [the new tests](https://github.com/taskcluster/generic-worker/blob/30714a5325b4bef926454ef197658fd2a09132b2/mounts_test.go#L311-L500) I've added today - to make sure all the logging is included that is required.

Can you check they contain everything you want, and that they look sound.

Thanks!

petemoore:
Failure is https://tools.taskcluster.net/groups/EZ2iuG6jS-6WLsC8yuc5NA/tasks/OT3WIUVBRyi7jctL30ylJQ/runs/0/logs/public%2Flogs%2Flive.log - I think unrelated.

petemoore:
@jhford I added SHA256 support for mounting content which can be from a plain URL or also from a task artifact. This will not be so useful for artifact content once we switch to the new blob storage type, but will still be useful for url content (when we mount a file from an arbitrary url in the task directory). When we switch over to using `taskcluster-lib-artifact-go` we could remove it here. It may be useful for mounting artifacts that are not in blob storage, or mounting from plain urls.

The main motivation of this PR is to get better logging - SHA256 checking was more of an aside. See [bug 1459376](https://bugzil.la/1459376) for more context.

petemoore:
@indygreg Also feel free to add review comments. I couldn't assign you a review, I think maybe you might need to be a member of the taskcluster org for that - but in any case, please feel free to add comments if there is anything you'd like me to address.

petemoore:
Note, I've created [bug 1460267](https://bugzil.la/1460267) about the gecko-t-win10-64-cu workers not starting up (for the CI). But we can see all the tests pass on all other platforms (including win2012 and win7) so since this change has nothing platform specific, I think we aren't blocked by the win10 CI not running yet.

mounts.go
```diff
@@ -492,54 +507,82 @@ func (f *FileMount) Mount() error {
...
+			SHA256:   sha256,
 		}
-	} else {
-		// Sanity check - if file is in file map, but not on file system,
-		// something is seriously wrong, so should be a worker exception
-		// (panic), not a task failure
-		_, err := os.Stat(fileCaches[cacheKey].Location)
-		if err != nil {
-			panic(fmt.Errorf("File in cache, but not on filesystem: %v", *fileCaches[cacheKey]))
+		err = validateSHA256(fsContent.String(), sha256, fsContent.RequiredSHA256(), task)
```
indygreg: Is this unhandled `err` result a bug?

petemoore: No, in this case, it is a [named result parameter](https://golang.org/doc/effective_go.html#named-results), which will is handled by calling code.

petemoore:
> I'm not fluent in Go, so I can't conduct a good code review. But the design seems reasonable and I really appreciate the logging, especially how the SHA256 of everything is logged everywhere. Helps with auditing.

> I couldn't tell from the code, but for large payloads, it will be a lot more efficient to simultaneously stream the download to a file and to a hasher (as opposed to downloading to memory and then performing a single large write and hash update). However, if you stream to a file, you need to be careful that the file is deleted if hash verification fails or another error occurs. The way I've implemented this elsewhere is to stream to a temporary file of the name <dest>.tmp and rename the file once the download is complete and verified.

We stream directly from http response into a file at [line 720 of mounts.go](https://github.com/taskcluster/generic-worker/pull/89/files#diff-ad1c67f33ec168b1fa498be41ca3c609R720), using an exponential backoff algorithm in the case of connection/download failures.

If the SHA256 does not match the required value, I see three possible cases this covers:

1) The value specified in the task definition simply wrong, and was incorrectly specified/calculated.
2) The value in the task definition correctly specifies the SHA256, but the download was corrupted, without the error detection machinery detecting this (e.g. connection is not dropped, no error is returned when reading from http response body, content length matches, http status code 200 is received, ....).
3) The value in the task definition is correct, but the url, for some reason, is returning different content.

Deleting the downloaded file would help in case 2, but I wonder how frequent this is. If there is an intermittent network issue, we should get an error reading from the stream, and would automatically retry, rather than reaching EOF. It implies bad bytes would be sent over the network, and that the Content-Length, if provided, was correct, so exactly the correct number of bytes were sent.

In case 1, I think it would be disadvantageous to delete the file, although only in terms of making the process less efficient, rather than causing any actual harm. This is a hit we could be prepared to take.

In case 3, deleting the content is probably overall a benefit too, because we are not caching something we don't care about.

Overall it probably does make sense to delete the file if the SHA256 is wrong - a question is whether we should handle this as part of the retry mechanics. We probably should. If it repeatedly results in the same unwanted content being downloaded, the number of retries is in any case bounded.

Should an incorrect SHA256 result in a task failure / task exception? Hard to know. If the retries consistently return invalid content, should the task be retried on a different worker?

What should we do if the SHA256 in the payload is different from the SHA256 of the file in the cache - should we delete and redownload?

I'll take a little bit of time to think about these things...

mounts.go
```diff
@@ -529,49 +534,57 @@ func (f *FileMount) Unmount(task *TaskRun) error {
...
-			task.Infof("[mounts] Found existing download for %v with (correct) SHA256 %v", cacheKey, sha256)
-		}
-	} else {
-		task.Warnf("[mounts] Found existing download for %v with SHA256 %v, but since no required SHA256 is specified in task mounts, it is not possible to verify the integrity of this file", cacheKey, sha256)
+	if requiredSHA256 == "" {
+		task.Warnf("[mounts] Content from %v has SHA256 %v but task payload does not declare a required value, so content authenticity cannot be verified", fsContent, sha256)
+		return
+	}
+	if requiredSHA256 != sha256 {
+		err = fmt.Errorf("SHA256 of content just downloaded from %v was %v but task definition explicitly requires %v so deleting it and giving up (only one attempt is made if there are no connection failures)", fsContent, sha256, requiredSHA256)
```
jhford: printing such a detailed error message looks great!  If it weren't a security risk, I'd love it if we could print out the first 50bytes in base64 to be able to do things like see that an S3 error message was downloaded instead of a binary file.  I don't think we should do this in public logs, though.

PR #91, opened 2018-05-17
Bug 1462369 - kill entire process tree on Windows when aborting a task command

petemoore:
See [bug 1462369](https://bugzil.la/1462369) for more details.


petemoore:
This failure is unrelated... I will look into that tomorrow.

PR #92, opened 2018-05-18
Move task reclaims into TaskStatusManager

petemoore:
This is a cleanup of the reclaim logic, to put it all in the `TaskStatusManager`. Callers don't need to concern themselves with reclaims, when you create the `TaskStatusManager`, it now starts the reclaims in a private (non-exported) background go routine. When a task is aborted or completes naturally, the reclaiming automatically stops.

### DIFF
--- a/taskstatus.go
+++ b/taskstatus.go
@@ -49,9 +49,9 @@ type TaskStatusManager struct {
 	statusChangeListeners map[*TaskStatusChangeListener]bool
 	abortException        *CommandExecutionError
 	// closed when reclaim go routine should stop reclaiming
-	stopReclaiming chan struct{}
+	stopReclaiming chan<- struct{}
 	// closed when reclaim loop exits
-	reclaimingDone chan struct{}
+	reclaimingDone <-chan struct{}
 	// true if reclaims are no longer taking place for this task
 	finishedReclaiming bool
 }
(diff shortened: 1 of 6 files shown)


petemoore:
TODO: In main go routine, I should have a `select` on waiting for `cmd.Wait()` to complete *or* for task to be aborted. Currently I just use `cmd.Wait()` and assume that if task is aborted, processes will be killed, and `cmd.Wait()` will return. If instead I don't assume processes will get killed and `cmd.Wait()` will successfully return, using a `select` statement will allow us to continue in either case of an abort or a task completion. Then process termination can be handled in a separate go routine while the worker resolves the aborted task and uploads artifacts etc.

Another option is just to only stop reclaiming once the task is about to be resolved in completed/failed/errored, not once it is aborted, since task abortion currently waits for processes to die so task claim could get expired.

taskstatus.go
```diff
@@ -278,10 +288,69 @@ func (tsm *TaskStatusManager) updateStatus(ts TaskStatus, f func(task *TaskRun)
...
 func NewTaskStatusManager(task *TaskRun) *TaskStatusManager {
-	return &TaskStatusManager{
+
+	tsm := &TaskStatusManager{
 		task:                  task,
 		takenUntil:            task.TaskClaimResponse.TakenUntil,
 		status:                task.TaskClaimResponse.Status,
 		statusChangeListeners: map[*TaskStatusChangeListener]bool{},
+		stopReclaiming:        make(chan struct{}),
+		reclaimingDone:        make(chan struct{}),
```
jonasfj: If you want, you can make the channel in a variable and assign it to the field... That way you can change the field type to receiving channel... It's just a nits... To prevent other methods from closing the channel...

PR #96, opened 2018-06-05
Prefix generated git branch names with X 

petemoore:
We do this since `taskId`s may start with `-`, for example, if retriggered using a tool that doesn't use _nice_ slugs.


petemoore:
Thanks!

PR #98, opened 2018-06-07
Bug 1439588 - support running whitelisted commands as Administrator

petemoore:
See [bug 1439588](https://bugzil.la/1439588) for context.


petemoore:
Hey John,

This PR is rather enormous, apologies! The individual commits might be easier to review than the whole thing. Let me know if you want me to split up the commits further to make them more understandable. Thanks in advance!

generated_windows.go
```diff
@@ -106,6 +106,14 @@ type (
 		// Since: generic-worker 5.3.0
 		ChainOfTrust bool `json:"chainOfTrust,omitempty"`
 
+		// Runs commands as Administrator. Property `task.payload.osGroups`
+		// must include `"Administrators"`. Requires scope
+		// `generic-worker:run-as-administrator:<provisionerId>/<workerType>`.
```
jhford: I'm curious why there's a separate scope here.  Is it to be extra certain that the command truly does want to run as administrator?  Would we ever include Administrators in the osGroups but not let them run as administrator?  It seems to be that osGroups is already scope-protected.

os_groups.go
```diff
@@ -42,25 +48,36 @@ func (osGroups *OSGroups) ReservedArtifacts() []string {
 func (osGroups *OSGroups) RequiredScopes() scopes.Required {
 	requiredScopes := make([]string, len(osGroups.Task.Payload.OSGroups), len(osGroups.Task.Payload.OSGroups))
 	for i, osGroup := range osGroups.Task.Payload.OSGroups {
-		requiredScopes[i] = "generic-worker:os-group:" + osGroup
+		requiredScopes[i] = "generic-worker:os-group:" + config.ProvisionerID + "/" + config.WorkerType + "/" + osGroup
```
jhford: nice change, specifying by worker type seems like the right thing to do here for sure.

plat_all-unix-style.go
```diff
@@ -136,11 +135,12 @@ func RenameCrossDevice(oldpath, newpath string) error {
 	return os.Rename(oldpath, newpath)
 }
 
-func (task *TaskRun) addGroupsToUser(groups []string) error {
-	if len(groups) == 0 {
-		return nil
-	}
-	return fmt.Errorf("Not able to add groups %v to user on platform %v - feature not supported", groups, runtime.GOOS)
+func (task *TaskRun) removeUserFromGroups(groups []string) (updatedGroups []string, notUpdatedGroups []string) {
+	return updatedGroups, groups
```
jhford: is it worthwhile to add a logging call to mention that this call (and one to addUserToGroups) when the updatedGroups aren't equivalent to groups?

process/misc_windows.go
```diff
@@ -0,0 +1,80 @@
+package process
```
jhford: does this need a windows build directive or do we only import this module in windows-specific code?

runasadministrator_windows.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (feature *RunAsAdministratorFeature) NewTaskFeature(task *TaskRun) TaskFeature {
+	return &RunAsAdministratorTask{
+		task: task,
+	}
+}
+
+func (l *RunAsAdministratorTask) RequiredScopes() scopes.Required {
+	return scopes.Required{{
```
jhford: Would it be worthwhile adding the os-group scope requirement as well?

runasadministrator_windows_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+func TestRunAsAdministratorDisabled(t *testing.T) {
+	defer setup(t)()
+	if config.RunTasksAsCurrentUser {
+		t.Skip("Skipping since running as current user...")
+	}
+	payload := GenericWorkerPayload{
+		Command: []string{
+			`whoami /groups`,
+			`whoami /groups | C:\Windows\System32\find.exe "S-1-16-12288" > nul`,
```
jhford: Could you add a comment which explains why this security id is being used?  I did some searching and found [this](https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems).  I'm not too sure how these identifiers work, but would it be worth while testing for the absence of "S-1-5-32-544" instead -- or is this security id the one that UAC would give to a ran-as-admin user?

artifacts_test.go
```diff
@@ -53,7 +53,7 @@ func validateArtifacts(
 
 func TestFileArtifactWithNames(t *testing.T) {
 
-	defer setup(t, "TestFileArtifactWithNames")()
+	defer setup(t)()
```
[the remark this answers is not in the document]
petemoore: Thanks! In earlier versions of go the `.Name()` method wasn't available, but now that it is, it is nice to get rid of this duplication. 😄

generated_all-unix-style.go
```diff
@@ -579,7 +580,7 @@ func taskPayloadSchema() string {
       "type": "object"
     },
     "osGroups": {
-      "description": "A list of OS Groups that the task user should be a member of. Requires\nscope ` + "`" + `generic-worker:os-group:\u003cos-group\u003e` + "`" + ` for each group listed.\n\nSince: generic-worker 6.0.0",
+      "description": "A list of OS Groups that the task user should be a member of. Requires scope\n` + "`" + `generic-worker:os-group:\u003cprovisionerId\u003e/\u003cworkerType\u003e/\u003cos-group\u003e` + "`" + ` for each\ngroup listed.\n\nSince: generic-worker 6.0.0",
```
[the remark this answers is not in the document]
petemoore: The yaml to json library I'm using seems to do this. Does seem a little unnecessary I agree!

generated_windows.go
petemoore: It is a great question! It turns out the `Administrators` group can be renamed, so there is no guarantee that if a group called `Administrators` exists, that it has administrative privileges. Funnily enough [Microsoft recommends renaming the Administrators group](https://technet.microsoft.com/en-us/library/cc700835.aspx?f=255&MSPPError=-2147217396#XSLTsection124121120120) (gotta love indirection, right?!).

Another subtle distinction is that new groups can be created with a subset of administrative privileges, so `runAsAdministrator` would in that case enable the administrative privileges available in the groups of the user, rather than all administrative privileges of the `Administrators` group, if it isn't a member of the Administrators group. In other words, as far as I understand, a user can still have Administrative privileges even without being a member of the `Administrators` group, or whatever it is renamed to.

We are planning to extend the UAC integration to allow a `sudo`-like feature where there is a file (similar to `/etc/sudoers` on linux/unix) that allows a whitelist of protected executables to be executed with administrative privileges. The `runAsAdministrator` feature is intended to be shortlived until the new `sudo`-like feature is available.

plat_all-unix-style.go
petemoore: That is logged in the calling methods, e.g. [here](https://github.com/taskcluster/generic-worker/blob/592fd32c28d16a1fd68a4a426a4b967f79a91359/os_groups.go#L64-L68).

process/misc_windows.go
petemoore: That's right - the name of the file tells the compiler to only include it in windows builds. See [go/build package docs](https://golang.org/pkg/go/build/#hdr-Build_Constraints).

runasadministrator_windows.go
petemoore: I think this is covered by [this comment](https://github.com/taskcluster/generic-worker/pull/98#discussion_r202306396).

runasadministrator_windows_test.go
petemoore: Yes, I'll do that, nice spot! :+1:

gwconfig/config.go
```diff
@@ -74,3 +80,48 @@ func (c *Config) String() string {
 	}
 	return string(json)
 }
+
+func (c *Config) Validate() error {
```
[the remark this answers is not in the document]
petemoore: Thanks. 😄

helper_test.go
```diff
@@ -72,7 +72,7 @@ func setupEnvironment(t *testing.T, testName string) (teardown func()) {
 
 	// useful for expiry dates of tasks
 	inAnHour = tcclient.Time(time.Now().Add(time.Hour * 1))
-	globalTestName = testName
+	globalTestName = t.Name()
```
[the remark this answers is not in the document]
petemoore: Thanks. 😄

os_groups.go
petemoore: Thanks. 😄

generated_windows.go
jhford: wow!  yah, I think separate scopes is good in that case.

plat_all-unix-style.go
jhford: great!

process/misc_windows.go
jhford: neat!  I didn't know that

runasadministrator_windows.go
jhford: yep, fully.

PR #102, opened 2018-06-16
Take Queue/PurgeCache/Auth Base URLs from provisioner

petemoore:
In preparation for [bug 12345](https://bugzil.la/12345), these changes take the Base URLs for

* Auth
* Purge Cache
* Queue

from User Data supplied by the AWS Provisioner.


djmitche:
++ to that!

PR #103, opened 2018-06-20
Bug 1469402 - Add a onExitStatus section to the payload

imbstack:
This is an attempt at adding `onExitStatus` like we have in docker-worker. If what I've done here mostly makes sense maybe you can give some pointers on testing that I should do and I'll try to do that tomorrow!

Note 1: This appears to work from running the worker locally and submitting test jobs.
Note 2: I'm not sure why the tests are failing? Something to do with `git status` on the mac builder at least.

### DIFF
--- a/all-unix-style.yml
+++ b/all-unix-style.yml
@@ -150,6 +150,22 @@ properties:
           [the github project](https://github.com/taskcluster/taskcluster-proxy) for more information.
 
           Since: generic-worker 10.6.0
+  onExitStatus:
+    title: Exit status handling
+    description: By default generic-worker will fail a task with a non-zero exit status
+      without retrying.  This payload property allows a task owner to define certain
+      exit statuses that will be marked as a retriable exception.
+    type: object
+    additionalProperties: false
+    properties:
+      retry:
+        title: Retriable exit statuses
+        description: If the task exists with a retriable exit status, the task will
+          be marked as an exception and a new run created.
+        type: array
+        items:
+          title: Exit statuses
+          type: integer
   mounts:
     type: array
     description: |-
--- a/all-unix-style.yml
+++ b/all-unix-style.yml
@@ -151,20 +151,23 @@ properties:
 
           Since: generic-worker 10.6.0
   onExitStatus:
-    title: Exit status handling
-    description: By default generic-worker will fail a task with a non-zero exit status
+    title: Exit code handling
+    description: By default generic-worker will fail a task with a non-zero exit code
       without retrying.  This payload property allows a task owner to define certain
-      exit statuses that will be marked as a retriable exception.
+      exit codes that are handled differently.
     type: object
     additionalProperties: false
     properties:
       retry:
-        title: Retriable exit statuses
-        description: If the task exists with a retriable exit status, the task will
-          be marked as an exception and a new run created.
+        title: Intermittent task exit codes
+        description: If the task exists with a retriable exit code, the task will
+          be marked as an exception with a reason of intermitten-task.
+          This may cause the queue to rerun the task.
+
+          See [itermittent tasks](https://docs.taskcluster.net/docs/reference/platform/taskcluster-queue/docs/worker-interaction#intermittent-tasks) for more detail.
         type: array
         items:
-          title: Exit statuses
+          title: Exit codes
           type: integer
   mounts:
     type: array
--- a/main.go
+++ b/main.go
@@ -946,7 +946,7 @@ func (err *CommandExecutionError) Error() string {
 	return fmt.Sprintf("%v", err.Cause)
 }
 
-func (task *TaskRun) RetryExitCode(c int64) bool {
+func (task *TaskRun) IsIntermittentExitCode(c int64) bool {
 	for _, code := range task.Payload.OnExitStatus.Retry {
 		if c == code {
 			return true
@@ -970,9 +970,9 @@ func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
 
 	switch {
 	case result.Failed():
-		if task.RetryExitCode(int64(result.ExitCode())) {
+		if task.IsIntermittentExitCode(int64(result.ExitCode())) {
 			return &CommandExecutionError{
-				Cause:      fmt.Errorf("Intermittent task will be rerun. Exit code was in onExitStatusList"),
+				Cause:      fmt.Errorf("Task appears to have failed intermittently - exit code %v found in task payload.onExitStatus list", result.ExitCode()),
 				Reason:     intermittentTask,
 				TaskStatus: errored,
 			}
--- a/main.go
+++ b/main.go
@@ -946,6 +946,15 @@ func (err *CommandExecutionError) Error() string {
 	return fmt.Sprintf("%v", err.Cause)
 }
 
+func (task *TaskRun) RetryExitCode(c int64) bool {
+	for _, code := range task.Payload.OnExitStatus.Retry {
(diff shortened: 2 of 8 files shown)


petemoore:
This looks awesome!

> Note 1: This appears to work from running the worker locally and submitting test jobs.

Yay!

> Note 2: I'm not sure why the tests are failing? Something to do with `git status` on the mac builder at least. 

That just means that the latest version of generated code hasn't been checked in. The CI runs the code generation, and if `git status` shows any locally modified files, it will fail the build. It looks the following changes made it into `windows.yml` but not into the generated code for Windows:

* Adding `"additionalProperties": false,`
* Renaming `requiredProperties` to `required`

If you regenerate the code and check in `generated_windows.go` that should fix that issue.

petemoore:
> If what I've done here mostly makes sense maybe you can give some pointers on testing that I should do and I'll try to do that tomorrow!

Have a look at `TestFailureResolvesAsFailure` in `main_test.go` as a starting point:

```go
 16 // Test failure should resolve as "failed"
 17 func TestFailureResolvesAsFailure(t *testing.T) {
 18     defer setup(t, "TestFailureResolvesAsFailure")()
 19     payload := GenericWorkerPayload{
 20         Command:    failCommand(),
 21         MaxRunTime: 10,
 22     }
 23     td := testTask(t)
 24 
 25     _ = submitAndAssert(t, td, payload, "failed", "failed")
 26 }
```

You could add some tests that also set the `onExitStatus` section of the task payload, and then use `failCommand()` as the command, and then check the task resolution just as the above test does.

It probably makes sense at this time to refactor this helper method, maybe rewrite `failCommand()` to `returnExitCode(exitCode uint)` or similar, and then update code that calls it to use the new name. Then any test you write can check for obscure exit codes etc.

Note `failCommand` is defined in both `helper_all-unix-style_test.go` and `helper_windows.go` to return appropriate commands for each platform.

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
...
+  onExitStatus:
+    title: Exit status handling
+    description: By default generic-worker will fail a task with a non-zero exit status
+      without retrying.  This payload property allows a task owner to define certain
+      exit statuses that will be marked as a retriable exception.
+    type: object
+    additionalProperties: false
+    properties:
+      retry:
+        title: Retriable exit statuses
```
petemoore: `title: Intermittent task exit codes`

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
...
+        description: If the task exists with a retriable exit status, the task will
+          be marked as an exception and a new run created.
```
petemoore:
```
description: Exit codes for any command in the task payload to cause the task to be resolved as `exception/intermittent-task`.
```

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
...
+        type: array
+        items:
+          title: Exit statuses
```
petemoore: Exit codes

main.go
```diff
@@ -946,6 +946,15 @@ func (err *CommandExecutionError) Error() string {
 	return fmt.Sprintf("%v", err.Cause)
 }
 
+func (task *TaskRun) RetryExitCode(c int64) bool {
```
petemoore: IsIntermittentExitCode

main.go
```diff
@@ -961,9 +970,17 @@ func (task *TaskRun) ExecuteCommand(index int) *CommandExecutionError {
 
 	switch {
 	case result.Failed():
-		return &CommandExecutionError{
-			Cause:      result.FailureCause(),
-			TaskStatus: failed,
+		if task.RetryExitCode(int64(result.ExitCode())) {
+			return &CommandExecutionError{
+				Cause:      fmt.Errorf("Intermittent task will be rerun. Exit code was in onExitStatusList"),
```
petemoore: Let's include the code and not mention rerun, as that depends on the Queue (whose behaviour may change to e.g. creating a retry rather than rerun in future).
```
Cause: fmt.Errorf("Task appears to have failed intermittently - exit code %v found in task payload.onExitStatus list", result.ExitCode()),
```

main_test.go
```diff
@@ -17,14 +17,65 @@ import (
...
 }
 
+// Exit codes specified in OnExitCode should resolve as itermittent
+func TestFailureRetryCode(t *testing.T) {
+	defer setup(t, "TestFailureResolvesAsFailure")()
+	payload := GenericWorkerPayload{
+		Command:    returnExitCode(123),
+		MaxRunTime: 10,
+		OnExitStatus: ExitStatusHandling{
+			Retry: []int64{123},
```
petemoore: Would be nice to have a list > 1 entry to make sure code iterates through all.

main_test.go
```diff
@@ -17,14 +17,65 @@ import (
...
+}
+
+// Exit codes _not_ specified in OnExitCode should resolve normally
+func TestFailureRetryCodeNormal(t *testing.T) {
+	defer setup(t, "TestFailureResolvesAsFailure")()
+	payload := GenericWorkerPayload{
+		Command:    returnExitCode(456),
+		MaxRunTime: 10,
+		OnExitStatus: ExitStatusHandling{
+			Retry: []int64{123},
```
petemoore: Again several codes here would be nice.

main_test.go
```diff
@@ -17,14 +17,65 @@ import (
...
+}
+
+// Exit codes should not override success
+func TestFailureRetryCodeSuccess(t *testing.T) {
+	defer setup(t, "TestFailureResolvesAsFailure")()
+	payload := GenericWorkerPayload{
+		Command:    returnExitCode(0),
+		MaxRunTime: 10,
+		OnExitStatus: ExitStatusHandling{
+			Retry: []int64{780},
```
petemoore: Again several codes here would be nice.

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
           [the github project](https://github.com/taskcluster/taskcluster-proxy) for more information.
 
           Since: generic-worker 10.6.0
+  onExitStatus:
```
petemoore: It is really unfortunate this is called `onExitStatus` and not `onExitCode` in docker-worker, as it is exit codes it deals with. But better to be consistent with docker-worker, just unfortunate...

Let's keep the name the same, but refer to exit codes in the doc, as that is the usual term.

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
...
+    title: Exit status handling
+    description: By default generic-worker will fail a task with a non-zero exit status
+      without retrying.  This payload property allows a task owner to define certain
+      exit statuses that will be marked as a retriable exception.
+    type: object
+    additionalProperties: false
+    properties:
+      retry:
```
petemoore: `exitCodes:` rather than `retry:`

all-unix-style.yml
```diff
@@ -150,6 +150,22 @@ properties:
           [the github project](https://github.com/taskcluster/taskcluster-proxy) for more information.
 
           Since: generic-worker 10.6.0
+  onExitStatus:
+    title: Exit status handling
```
petemoore: `title: Resolve task as intermittent`

all-unix-style.yml
imbstack: Note for others reading this, we think we're going to land this with a clone of docker-worker's behavior and then roll out this nicer form later.

petemoore:
Thanks Brian!

PR #105, opened 2018-06-29
Bug 1471541 - new tooltool url

garbas:


garbas:
is this a transient failure?

imbstack:
I reran the test and it appears to have failed the same way the second time. Not sure that it means it is caused by your change though.

petemoore:
These files are generated, and aren't used in production, they are part of an experiment to convert OCC manifests into plain powershell, but we don't use them anywhere.

I'll update the docs to say that, and update the generation code to add a comment to the generated powershell to say that the code is generated.

PR #106, opened 2018-07-17
Added utility scripts for testing new generic-worker releases on try

petemoore:
Hey Brian,

I took a first pass at moving these scripts into the repository. It isn't beautiful and the end state I'd like it to be in, but it unblocks other people from trying these things out.

I haven't gone heavy on docs, because I really want to rewrite the whole thing, probably in go, and ideally have it triggered via a git interaction of some kind. I haven't worked out the place for this yet, but my feeling is we should be using some kind of existing access controls (whether those are github access controls or taskcluster or some other system access controls I'm not sure) but some simple interaction should be able to select a generic-worker release (default being latest) and then choose to make a test gecko push with it, or nss, or both. On the way, sensible defaults, but having the option to adjust try syntax of those pushes, or select specific gecko/nss commits to be the basis of the try pushes, etc. Then when it all runs, it should run inside taskcluster so we have logs etc so people can monitor how it is progressing.

But step one was just getting the scripts into the repo that I was using so other people can easily use them too...

### DIFF
--- a/mozilla-try-scripts/gecko-try.sh
+++ b/mozilla-try-scripts/gecko-try.sh
@@ -1,0 +1,145 @@
+#!/bin/bash -e
+
+######
+# This script allows you to test a new generic-worker Windows release on
+# gecko try. It deploys your chosen generic-worker release to a set of
+# staging worker types that are created for this purpose, in order not
+# to impact other worker types while testing a generic-worker release.
+#
+# TODO: this should be rewritten as e.g. a go program, to reduce number
+# of system dependencies (e.g. bash, curl, jq, grep, sleep, ...).
+#
+# TODO: alternatively, it could probably be written to run in docker, or
+# as a taskcluster task that runs based on some github event.
+#
+######
+
+function open_browser_page {
+  case "$OSTYPE" in
+    linux*)
+      xdg-open "${1}"
+      ;;
+    darwin*)
+      open "${1}"
+      ;;
+  esac
+}
+
+NEW_VERSION="${1}"
+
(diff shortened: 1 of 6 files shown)


mozilla-try-scripts/gecko-try.sh
```diff
@@ -0,0 +1,144 @@
...
+
+NEW_VERSION="${1}"
+
+if [ -z "${NEW_VERSION}" ]; then
+  echo "Please specify version of generic-worker to use in gecko try push, e.g. '${0}' 10.4.1" >&2
+  exit 64
+fi
+
+echo "Checking system dependencies..."
+for command in curl jq grep sleep file git hg cat sed rm mktemp python go; do
```
imbstack: Awesome.

I think this list isn't particularly onerous either. Most Mozilla/Taskluster people at least will have all of these installed.

mozilla-try-scripts/gecko-try.sh
```diff
@@ -0,0 +1,144 @@
...
+
+cd ~/.mozilla-central
+hg up -C
+hg purge
+
+# wait for OCC deployment to complete
+go run "${THIS_SCRIPT_DIR}/waitforOCC.go" "${OCC_COMMIT}"
+
+hg pull -u -r default
+curl -L 'https://bug1400012.bmoattachments.org/attachment.cgi?id=8948627' | hg import -
```
imbstack: Can we check this in to the repo directly instead of curling it?

petemoore: Good point... I'll do that!

PR #107, opened 2018-07-18
Bug 1476441 - fix docs.taskcluster.net links

djmitche:


djmitche:
@petemoore ping

djmitche:
There are plans to eventually move this documentation out of the README, but for now we're just dealing with the broken links on github.  They can't work both places -- all we can do is not have the links in the README.

PR #114, opened 2018-07-27
Make defaults for CachesDir and DownloadsDir a relative path

### LINKED ISSUE #109: cachesDir and downloadsDir default to a Windows path on unix
They default, respectively, to:
```
"C:\\generic-worker\\caches"
```
and
```
"C:\\generic-worker\\downloads"
```

petemoore:
This should fix #109. Note, in our current configuration, we were running generic-worker on windows with `C:\generic-worker` as the working directory, so using a relative path for the caches and downloads directories won't change anything on our windows workers, but having a relative path will mean sensible defaults for other platforms. Note these are only defaults, so workers can use different directories if they prefer.

### DIFF
--- a/main.go
+++ b/main.go
@@ -167,8 +167,11 @@ and reports back results to the queue.
 
           authBaseURL                       The base URL for API calls to the auth service.
           availabilityZone                  The EC2 availability zone of the worker.
-          cachesDir                         The location where task caches should be stored on
-                                            the worker. [default: C:\generic-worker\caches]
+          cachesDir                         The directory where task caches should be stored on
+                                            the worker. The directory will be created if it does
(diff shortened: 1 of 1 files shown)


petemoore:
CC @glandium

petemoore:
> I can't tell if the test fails are related or not?

The test failures are due to a bug in generic-worker 10.11.1alpha1 which was running on the Windows Server 2012 R2 worker type we use for the CI (win2012r2-cu). I've rolled back to 10.10.0 on the CI and will retrigger CI with an empty commit and some retriggers...

petemoore:
No problems after updating `generic-worker.exe` on `win2012r2-cu`, so merging!

PR #119, opened 2018-08-06
Bug 1480412 - allow empty osGroups list on non-Windows platforms

petemoore:
See [bug 1480412](https://bugzil.la/1480412) for more details.

Since `osGroups` feature isn't supported on non-Windows platforms currently, in the most recent release (10.11.2) I had moved the `osGroups` code into the Windows codebase (`*_windows.go` files) from the cross platform go files. I hadn't realised that the gecko task generation code was creating non-Windows tasks specifying an empty `osGroups` list (`[]`). Rather than fix the task generation in all gecko trees/branches (and any other projects that might be doing the same) not to allow `osGroups` for platforms where this feature is not implemented, this PR ensures that the only acceptable value for `osGroups` on a non-Windows platform is the empty list (i.e. "no groups").

The reason it isn't implemented on those other platforms is that the tasks run as the same user as the worker process. Since the user cannot grant itself more privileges than it already has, this feature only makes sense on platforms where the task runs as a different user to the worker process, which currently is only Windows.

Specifying a non-empty list will result in a malformed-payload exception.

This PR is mostly the undoing of the previous refactor, i.e. moving the code from windows-only go files (`*_windows_*.go` files) back to the cross platform (`*.go`) files and a couple of tests to ensure it works.

### DIFF
--- a/all-unix-style.yml
+++ b/all-unix-style.yml
@@ -159,6 +159,18 @@ properties:
     items:
       title: Mount
       "$ref": "#/definitions/mount"
+  osGroups:
+    type: array
+    title: OS Groups
+    description: |-
+      A list of OS Groups that the task user should be a member of. Not yet implemented on
+      non-Windows platforms, therefore this optional property may only be an empty array if
+      provided.
+
+      Since: generic-worker 6.0.0
+    items:
+      type: string
+    maxItems: 0
   supersederUrl:
     type: string
     title: Superseder URL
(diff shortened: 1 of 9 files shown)


petemoore:
@imbstack Note, all the failures are due to [bug 1481178](https://bugzil.la/1481178).

davehouse:
Thanks @petemoore for fixing this.

PR #120, opened 2018-09-07
Support GCE as an environment

imbstack:
This _works_ but might not be the way we want to do it in the future. Please feel free to have me change how anything works.

This works in concert with https://github.com/imbstack/taskcluster-gce-provider-concept.

Specifically the tc creds comes from [here](https://github.com/imbstack/taskcluster-gce-provider-concept/blob/master/src/api.js)

We can wait on merging this until rootUrl support exists.

### DIFF
--- a/main.go
+++ b/main.go
@@ -37,6 +37,8 @@ var (
 	cwd = CwdOrPanic()
 	// Whether we are running under the aws provisioner
 	configureForAws bool
+	// Whether we are running in GCP
+	configureForGcp bool
 	// General platform independent user settings, such as home directory, username...
 	// Platform specific data should be managed in plat_<platform>.go files
 	taskContext = &TaskContext{}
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -82,6 +84,7 @@ and reports back results to the queue.
   Usage:
     generic-worker run                      [--config         CONFIG-FILE]
                                             [--configure-for-aws]
+                                            [--configure-for-gcp]
```
petemoore: I guess we could make these two options mutually exclusive, e.g.

```
[[--configure-for-aws]|[--configure-for-aws]]
```

I'm not too hot on docopt syntax - it might be subtly different. Not a huge gain to be had by doing this, but maybe solves some problems of non-determinism if both were passed in?

djmitche:
I rebased and fixed the tests.

main.go
petemoore: This still needs doing.

djmitche:
Now blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1510052

djmitche:
OK, this passes tests for me locally, and adds a new test using a GCP mock.  I only copied one of the tests in aws_test.go, since the others seem not specific to AWS.

djmitche:
I'm not sure why taskcluster-proxy fails on amd64?  I don't think any of my changes would affect that..

djmitche:
Oh, tc-proxy is failing because credentials aren't available to pull requests from other repos.  Travis still hasn't figured that feature out, but TC has :)

PR #121, opened 2018-09-10
Bug 1434943 - Update MozillaBuild version to 3.2 for NSS

martinthomson:
We need to run 3.2 if we're going to use the vswhere tool to find VS.  That won't change the calculus for these builds, but it's a big usability improvement for contributors.


martinthomson:
Oh, I should point out that this shouldn't be merged until we have merged https://phabricator.services.mozilla.com/D5483 on the NSS side.  Otherwise we won't be able to find nsinstall - the reason we stayed with 3.1.1 in the first place.

PR #123, opened 2018-09-20
Bug 1469614 - work in progress

petemoore:
With this PR, `generic-worker` has new config property `taskclusterRootURL` in order to be interoperable with different taskcluster deployments.

Work in progress, since this depends on the client updates (i.e. depends on taskcluster/taskcluster-client-go#37).

### DIFF
--- a/aws_helper_test.go
+++ b/aws_helper_test.go
@@ -95,7 +95,7 @@ func (m *MockAWSProvisionedEnvironment) Setup(t *testing.T) func() {
 				"price":               3.02,
 				"launchSpecGenerated": time.Now(),
 				"lastModified":        time.Now().Add(time.Minute * -30),
-				"provisionerBaseUrl":  "http://localhost:13243/provisioner",
+				"taskclusterRootUrl":  "http://acme.com:43243/demo",
 				"securityToken":       "12345",
 			}
 			WriteJSON(t, w, resp)
--- a/main.go
+++ b/main.go
@@ -25,7 +25,6 @@ import (
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 	"github.com/taskcluster/taskcluster-client-go/tcauth"
 	"github.com/taskcluster/taskcluster-client-go/tcawsprovisioner"
-	"github.com/taskcluster/taskcluster-client-go/tcpurgecache"
 	"github.com/taskcluster/taskcluster-client-go/tcqueue"
 	"github.com/xeipuuv/gojsonschema"
 )
@@ -173,7 +172,6 @@ and reports back results to the queue.
         ** OPTIONAL ** properties
         =========================
 
-          authBaseURL                       The base URL for API calls to the auth service.
           availabilityZone                  The EC2 availability zone of the worker.
           cachesDir                         The directory where task caches should be stored on
                                             the worker. The directory will be created if it does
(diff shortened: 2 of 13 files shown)


aws_helper_test.go
```diff
@@ -95,7 +95,7 @@ func (m *MockAWSProvisionedEnvironment) Setup(t *testing.T) func() {
 				"price":               3.02,
 				"launchSpecGenerated": time.Now(),
 				"lastModified":        time.Now().Add(time.Minute * -30),
-				"provisionerBaseUrl":  "http://localhost:13243/provisioner",
+				"taskclusterRootUrl":  "http://acme.com:43243/demo",
```
djmitche: rootUrl's can't have a path (not that it really matters here, but fyi)

main.go
```diff
@@ -293,6 +286,8 @@ and reports back results to the queue.
                                             [default: taskcluster-proxy]
           taskclusterProxyPort              Port number for taskcluster-proxy HTTP requests.
                                             [default: 80]
+          taskclusterRootURL                Root URL of taskcluster cluster that this worker
+                                            should interact with. [default: https://taskcluster.net]
```
djmitche: I think a default is OK here temporarily, but once everything that launches workers is specifying a rootUrl, let's plan to remove the default.

mounts.go
```diff
@@ -151,8 +151,7 @@ func (cm *CacheMap) LoadFromFile(stateFile string, cacheDir string) {
 func (feature *MountsFeature) Initialise() error {
 	fileCaches.LoadFromFile("file-caches.json", config.CachesDir)
 	directoryCaches.LoadFromFile("directory-caches.json", config.DownloadsDir)
-	pc = tcpurgecache.New(nil)
-	pc.BaseURL = config.PurgeCacheBaseURL
+	pc = tcpurgecache.New(&tcclient.Credentials{RootURL: config.TaskclusterRootURL})
```
djmitche: This is a good example of why I suggest that the client's `New` fail if `RootUrl` isn't provided -- unauthenticated requests like this would otherwise continue to work, and in this case at least not even cause any visible failures (workers in any cluster would just consult the https://taskcluster.net purge-cache instance -- until the day we shut that off).

The automatic use of env vars by the client makes the risk a little worse, too -- I expect `tcpurgecache.New(nil)` would work as long as TASKCLUSTER_ROOT_URL was set, which it probably is during testing but probably is not in production.  In the JS tests, this caused some unhappiness, and we now [unset the environment variables](https://github.com/taskcluster/taskcluster-lib-testing/commit/05023065ea1435f95c789ab6d40a4b6079ca1f40) after reading them into the config.  So that's something to watch out for.

petemoore:
This PR got superseded by #139 after this bit-rotted a little, and I decided a fresh approach was better than updating this PR.

In #139, I removed the default value for the taskcluster root URL, so it does not need to be specified in the worker config file.

Note, root URL and base URL are complementary concepts that apply at different levels. The root URL applies to a taskcluster deployment, whereas base URL applies to a single service. We configure the cluster using root URL, but internally the individual service clients store their base URL. This is key, because it decouples the concerns of an individual service client from the concerns of a full taskcluster deployment, which is a concern of a higher level.

For example, we have generic-worker integration tests which mock the AWS Provisioner. These tests need to be able to swap out the actual provisioner client for the mock provisioner, which they can do, because the mock provisioner base URL can be adjusted to point to a localhost URL, while keeping all other service clients pointing at the environment used for the integration testing. This wouldn't be possible if internally we didn't associate a service client with a base URL.

Base URL concerns the individual service client, and is _complementary_ to the root URL. The base URL is typically derived from the root URL, but may be adjusted, for example when mocking a service. But under normal operation, clients will be configured from the root URL, so this is transparent to the user unless they actively change it.

PR #124, opened 2018-10-01
Bug 1495435 - Make --configure-for-aws a configurable option at install time

petemoore:
See [bug 1495435](https://bugzilla.mozilla.org/show_bug.cgi?id=1495435).

Previously, the generic-worker `install` target on Windows created a `run-generic-worker.bat` script that called `generic-worker run --configure-for-aws`, i.e. it made the assumption that the worker ran in AWS.

This PR adds the existing `--configure-for-aws` option of `generic-worker run` to `generic-worker install`. If (and only if) `--configure-for-aws` is specified when calling `generic-worker install`, the generated `run-generic-worker.bat` script will include `--configure-for-aws` in the `generic-worker run` command.

In other words, you now choose at _install_ time if this installation is for AWS workers, and the correct option is passed down to the `generic-worker run` command.

### DIFF
--- a/main.go
+++ b/main.go
@@ -86,6 +86,7 @@ and reports back results to the queue.
     generic-worker install service          [--nssm           NSSM-EXE]
                                             [--service-name   SERVICE-NAME]
                                             [--config         CONFIG-FILE]
+                                            [--configure-for-aws]
     generic-worker show-payload-schema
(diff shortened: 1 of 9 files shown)


petemoore:
@SimonSapin FYI

PR #127, opened 2018-11-02
Bug 1479415 - tasks executing a non-executable file should resolve as failed, not exception

petemoore:
See [bug 1479415](https://bugzil.la/1479415) for context.

### DIFF
--- a/process/process_all-unix-style.go
+++ b/process/process_all-unix-style.go
@@ -31,23 +31,51 @@ type Result struct {
 }
 
 func (r *Result) Succeeded() bool {
-	return r.SystemError == nil && r.ExitError == nil
+	return r.SystemError == nil && r.ExitError == nil && !r.Aborted
 }
 
+// Unlike on Windows, a system error is grounds for a task failure, rather than
+// a task exception, since it can be caused by e.g. a task trying to execute a
+// command that doesn't exist, or trying to execute a file that isn't
+// executable. Therefore a system error, or an exit error (process ran but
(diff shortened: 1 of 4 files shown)


process/process_all-unix-style.go
```diff
@@ -31,23 +31,51 @@ type Result struct {
...
+	}
+	if r.SystemError != nil {
+		return r.SystemError
+	}
+	return nil
 }
 
+// Unlike on Windows, if there is a system error, we don't crash the worker,
+// since this can be caused by task that tries to execute a non-existing
+// command or a file that isn't executable. On Windows all commands are
```
jhford: Is this an exhaustive list of cases we care about?  If so, if there's an error, we could do an os.Stat on argv[0] to see if it exists and if it has execute permissions and still panic for other errors.  If not, then ignore my comment :)

petemoore: It is a good question! I can imagine there could be other cases, such as the file exists, has the executable bit set, but isn't a valid executable. Perhaps this would result in a bad exit code, or perhaps a system error when executing it, I'm not sure. I figured in this grey area where there could be cases we haven't thought about, better to err on the side of caution and resolve as a failure, rather than shutting down the worker, as if a task can be constructed that causes a system error here, this could be potentially used as an attack vector for a malicious task author to create a DoS attack against a worker pool, causing all of the workers to shut down. This is perhaps preferable to dealing with failures that should have been an exception. In general, since we're running untrusted code, I think better to assume the task is at fault if we're not really sure.

PR #128, opened 2018-11-08
Add godep to generic-worker

walac:
godep [1] handles dependency management for go packages. It creates a
vendor/ subdirectory with all packages this repository depends on.

Notice golint will report a lot more errors as the current version
doesn't support an option to ignore the vendor/ directory yet.

[1] https://github.com/golang/dep


.taskcluster.yml
```diff
@@ -101,7 +101,7 @@ tasks:
           exit /b 0
         - git rev-parse HEAD > revision.txt
         - set /p revision=< revision.txt
-        - go get -ldflags "-X main.revision=%revision%" -v -u -t ./...
+        - go install -ldflags "-X main.revision=%revision%"
```
petemoore: We still need `./...` to ensure the command line tools are built, and I'd like to keep the `-v` for getting more feedback about the build process.

.taskcluster.yml
```diff
@@ -173,7 +173,7 @@ tasks:
         - cd ..
         - go generate
         - set revision={{ event.head.sha }}
-        - go get -ldflags "-X main.revision=%revision%" -v -u -t ./...
+        - go install -ldflags "-X main.revision=%revision%"
```
petemoore: Same here, let's keep `-v ./...`.

.taskcluster.yml
```diff
@@ -254,7 +254,7 @@ tasks:
           exit /b 0
         - git rev-parse HEAD > revision.txt
         - set /p revision=< revision.txt
-        - go get -ldflags "-X main.revision=%revision%" -v -u -t ./...
+        - go install -ldflags "-X main.revision=%revision%"
```
petemoore: Same here.

.taskcluster.yml
```diff
@@ -328,7 +328,7 @@ tasks:
             go get -v -u
             cd ..
             go generate
-            go get -ldflags "-X main.revision=$(git rev-parse HEAD)" -v -u -t ./...
+            go install -ldflags "-X main.revision=$(git rev-parse HEAD)"
```
petemoore: Same here.

.travis.yml
```diff
@@ -20,7 +20,6 @@ script:
   - "test $GIMME_OS.$GIMME_ARCH != linux.amd64 || go generate"
   - "test $GIMME_OS.$GIMME_ARCH != linux.amd64 || git status"
   - "test $GIMME_OS.$GIMME_ARCH != linux.amd64 || test $(git status --porcelain | wc -l) == 0"
-  - "go get -ldflags \"-X main.revision=$(git rev-parse HEAD)\" -v -t ./..."
```
petemoore: We still need to `go install` otherwise releases won't work. So please change to `go install -ldflags ... -v ./...`.

build.cmd
```diff
@@ -10,7 +10,6 @@ cd gw-codegen
 go get -v || exit /b %ERRORLEVEL%
 cd ..
 go generate || exit /b %ERRORLEVEL%
-go get -v -t ./... || exit /b %ERRORLEVEL%
```
petemoore: `go install -v ./... || exit /b %ERRORLEVEL%`

build.sh
```diff
@@ -48,13 +48,11 @@ go generate ./...
 
 function install {
   if [ "${1}" != 'native' ]; then
-    GOOS="${1}" GOARCH="${2}" CGO_ENABLED=0 go get -ldflags "-X main.revision=$(git rev-parse HEAD)" -v ./...
```
petemoore: Same here, let's use `go install` so we get the binaries built under `"${GOPATH}/bin"`.

```
GOOS="${1}" GOARCH="${2}" CGO_ENABLED=0 go install -ldflags "-X main.revision=$(git rev-parse HEAD)" -v ./...
```

build.sh
```diff
@@ -48,13 +48,11 @@ go generate ./...
...
     # GOOS="${1}" GOARCH="${2}" go vet ./...
     # note, this just builds tests, it doesn't run them!
     GOOS="${1}" GOARCH="${2}" CGO_ENABLED=0 go test -c github.com/taskcluster/generic-worker
     GOOS="${1}" GOARCH="${2}" CGO_ENABLED=0 go test -c github.com/taskcluster/generic-worker/livelog
   else
-    CGO_ENABLED=0 go get -ldflags "-X main.revision=$(git rev-parse HEAD)" -v ./...
```
petemoore: And here, please put it back and change `get` to `install`.

djmitche:
I would love to see this landed, as otherwise I get issues like [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1510052) just about every time I drop by to try to build generic-worker.

djmitche:
Somewhat related, we can't do deterministic builds in `taskcluster-builder` without this support (as those always build from a clean source tree).

djmitche:
:tada:

PR #134, opened 2018-12-08
remove 'git clean' command from build process

djmitche:
We are not sure why this was here - perhaps a copy-paste error?  But it
deletes local files that I would like it not to delete!


winstyfreddy:
Many thanks!

PR #135, opened 2018-12-28
Bug 1516458 - take rootURL as config and set TASKCLUSTER_ROOT_URL and TASKCLUSTER_PROXY_URL

djmitche:
For tests, we just run `go get github.com/taskcluster/taskcluster-proxy, so this uses the newest version.  I'm not sure how to accomplish that in the prod deployment.


djmitche:
hm, well, this isn't working yet..

djmitche:
(somehow the tests passed for me locally once, but not anymore)

djmitche:
OK, I think I figured it out -- `generateCommand` is called before the features are initialized, so the `Env` property of the task payload has already been read at feature initialization time.  I switched to modifying each `exec.Cmd`'s `.Env` property directly, as suggested by

```go
    // generate commands, in case features want to modify them
    for i := range task.Payload.Command {
        err := task.generateCommand(i) // platform specific                                                                                                                                                                                                                                 
        if err != nil {
            panic(err)
        }    
    }    
```

This fixes the tests for me locally.  The error in CI on Windows is

```
C:\Users\task_1546265954\gopath1.10.3\src\github.com\taskcluster\generic-worker\testdata\TestTaskclusterProxy\task_1546265954>go run "curlget.go" $TASKCLUSTER_PROXY_URL/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt 
2018/12/31 14:28:57 Get http://localhost:34569/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt: dial tcp: lookup localhost: getaddrinfow: A non-recoverable error occurred during a database lookup.
exit status 1
```
which confuses me -- the PR makes the following change to the URL:
```
-                               fmt.Sprintf("http://localhost:%v/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt", config.TaskclusterProxyPort),
+                               fmt.Sprintf("$TASKCLUSTER_PROXY_URL/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt"),
```

so both before and (from the logging) after, it's `http://localhost:<port>/...`.  Why would it suddenly have trouble looking up `localhost`?

djmitche:
The Travis build is failing because env vars aren't available to PRs

djmitche:
The same TestTaskclusterProxy test that's failing here seems to be failing reliably on master, just differently:

https://taskcluster-artifacts.net/flKO8uO0T3yA5rfyfQe2fg/0/public/logs/live_backing.log
https://taskcluster-artifacts.net/FbZXyJP7TDSirPQLhqdaIQ/0/public/logs/live_backing.log
https://taskcluster-artifacts.net/EGzzMzPNR6-gF9QiW5p9Lg/0/public/logs/live_backing.log
```
--- FAIL: TestTaskclusterProxy (67.03s)
	helper_test.go:211: Scheduled task b9dUhBvITfWfhHwCVfMPCQ
	helper_test.go:224: Something went wrong executing worker - got exit code 69 but was expecting exit code 0
```
(note when looking at those logs, there are two different test cases named TestTaskclusterProxy)

..so perhaps this has never worked, and my patch is unrelated?

petemoore:
> The same TestTaskclusterProxy test that's failing here seems to be failing reliably on master, just differently:
> 
> https://taskcluster-artifacts.net/flKO8uO0T3yA5rfyfQe2fg/0/public/logs/live_backing.log
> https://taskcluster-artifacts.net/FbZXyJP7TDSirPQLhqdaIQ/0/public/logs/live_backing.log
> https://taskcluster-artifacts.net/EGzzMzPNR6-gF9QiW5p9Lg/0/public/logs/live_backing.log
> 
> ```
> --- FAIL: TestTaskclusterProxy (67.03s)
> 	helper_test.go:211: Scheduled task b9dUhBvITfWfhHwCVfMPCQ
> 	helper_test.go:224: Something went wrong executing worker - got exit code 69 but was expecting exit code 0
> ```
> 
> (note when looking at those logs, there are two different test cases named TestTaskclusterProxy)
> 
> ..so perhaps this has never worked, and my patch is unrelated?

It looks like it may be related to your patch. See

https://tools.taskcluster.net/groups/WtVE-HaNSjW2zOvFQ2veGQ/tasks/b9dUhBvITfWfhHwCVfMPCQ/runs/0/logs/public%2Flogs%2Flive.log#L27-59

petemoore:
I suspect changes to the taskcluster proxy test(s) will be needed in light of https://github.com/taskcluster/taskcluster-proxy/pull/38

petemoore:
Although the comments in that PR say "This should be 100% backward compatible" so I'm not sure. Looking deeper now...

djmitche:
This PR contains the fixes to make generic-worker work with the latest tc-proxy (after 
https://github.com/taskcluster/taskcluster-proxy/pull/38).  But since generic-worker does not use a pinned version of the proxy, it makes sense that landing the tc-proxy PR broke CI for generic-worker until this is landed.

So we're back to localhost not resolving on windows.  Googling suggests that's because %SYSROOT% isn't set, and from the look of the logs, it is not set:

https://taskcluster-artifacts.net/NIlDOfEvR5msuvQlCi03gw/0/public/logs/live_backing.log
```
Z:\task_1546277086\gopath1.10.3\src\github.com\taskcluster\generic-worker\testdata\TestTaskclusterProxy\task_1546277086>set
COMSPEC=C:\windows\system32\cmd.exe
GOPATH=Z:\task_1546277086\gopath1.10.3
GOROOT=Z:\task_1546277086\go1.10.3\go
PATH=Z:\task_1546277086\git\bin;Z:\task_1546277086\gopath1.10.3\bin;Z:\task_1546277086\go1.10.3\go\bin;C:\Program Files\Mercurial;C:\mozilla-build\7zip;C:\mozilla-build\info-zip;C:\mozilla-build\kdiff3;C:\mozilla-build\moztools-x64\bin;C:\mozilla-build\mozmake;C:\mozilla-build\msys\bin;C:\mozilla-build\msys\local\bin;C:\mozilla-build\nsis-3.0b3;C:\mozilla-build\nsis-2.46u;C:\mozilla-build\python;C:\mozilla-build\python\Scripts;C:\mozilla-build\python3;C:\mozilla-build\upx391w;C:\mozilla-build\wget;C:\mozilla-build\yasm;C:\windows\system32;C:\windows;C:\windows\System32\Wbem;C:\windows\System32\WindowsPowerShell\v1.0\;C:\mozilla-build\python27;C:\mozilla-build\python27\Scripts;C:\mozilla-build\vim\vim72;C:\CoreUtils\bin;C:\mozilla-build\buildbotve\scripts;c:\Program Files\Microsoft Windows Performance Toolkit\;C:\Users\cltbld\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0;C:\ProgramData\chocolatey\bin;C:\Program Files\Puppet Labs\Puppet\bin;C:\mozilla-build\hg;C:\Program Files\GNU\GnuPG\pub;C:\Program Files\Mercurial\;
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.JS;.WS;.MSC
PROMPT=$P$G
TASKCLUSTER_PROXY_URL=http://localhost:34569
TASKCLUSTER_ROOT_URL=https://taskcluster.net
TASK_ID=NIlDOfEvR5msuvQlCi03gw
tcexitcode=0
```

docs/features.md
```diff
@@ -59,53 +59,35 @@ References:
...
+*NOTE*: as a special case, the scopes required to call
+`queue.createArtifact(<taskId>, <runId>, ..)` are automatically included,
+regardless of `task.scopes`.
 
-For example (using curl) inside a task container.
+The proxy is easy to use within a shell command, too:
 
 ```sh
-cat secret | curl --header 'Content-Type: application/json' --request PUT --data @- http://localhost/secrets/v1/secret/<secretName>
+curl $TASKCLUSTER_PROXY_URL/api/secrets/v1/secret/my-top-secret-secret
```
petemoore: I prefer the example with PUT as it demonstrates that the proxy is not only for fetching data, but for all taskcluster endpoints.

taskcluster_proxy.go
```diff
@@ -61,7 +65,7 @@ func (l *TaskclusterProxyTask) Start() *CommandExecutionError {
 			Certificate: l.task.TaskClaimResponse.Credentials.Certificate,
 			ClientID:    l.task.TaskClaimResponse.Credentials.ClientID,
 		},
-		l.task.TaskID,
+		scopes,
```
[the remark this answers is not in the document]
djmitche: They aren't actually authorizedScopes in this case -- I noticed it was done that way, but that's really just "cheating" and keeping the list of scopes in an otherwise-unused field.

taskcluster_proxy.go
```diff
@@ -41,17 +49,21 @@ func (l *TaskclusterProxyTask) ReservedArtifacts() []string {
...
-	}
-}
-
 func (l *TaskclusterProxyTask) RequiredScopes() scopes.Required {
 	// let's not require any scopes, to be consistent with docker-worker
 	return scopes.Required{}
 }
 
+func (l *TaskclusterProxyTask) setupEnvVars() {
+	// by this point, task.task.Payload.Env has already been used to create task.Commands, so
```
petemoore: `task.task.Payload.Env` -> `l.task.Payload.Env` ?

taskcluster_proxy_test.go
```diff
@@ -7,16 +7,23 @@ import (
...
 		Command: append(
 			append(
 				goEnv(),
+				// long enough to reclaim and get new credentials
 				sleep(12)...,
 			),
 			goRun(
 				"curlget.go",
-				fmt.Sprintf("http://localhost:%v/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt", config.TaskclusterProxyPort),
+				fmt.Sprintf("$TASKCLUSTER_PROXY_URL/queue/v1/task/KTBKfEgxR5GdfIIREQIvFQ/runs/0/artifacts/SampleArtifacts/_/X.txt"),
```
petemoore: This warrants a comment that this is literally replaced by `curlget.go` program, otherwise it is confusing, since there is no shell involved.

I would have preferred a more general solution like using `${TASKCLUSTER_PROXY_URL}` and then in `curlget.go` expanding all occurrences of `${...}` using the process environment.

tcproxy/tcproxy_test.go
```diff
@@ -10,7 +10,13 @@ import (
 	tcclient "github.com/taskcluster/taskcluster-client-go"
 )
 
-func TestTaskclusterProxy(t *testing.T) {
+func TestTcProxy(t *testing.T) {
```
petemoore: Why rename from `TestTaskclusterProxy` to `TestTcProxy`? I prefer the longer name.

testdata/curlget.go
```diff
@@ -5,13 +5,16 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strings"
 )
 
 func main() {
 	if len(os.Args) != 2 {
 		log.Fatal("Usage: go run curlget.go <url>")
```
petemoore: The Usage should mention that `$TASKCLUSTER_PROXY_URL` is (or env vars of the form `${`_env_var_`}` are) expanded.

taskcluster_proxy.go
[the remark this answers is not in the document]
petemoore: They really are AuthorizedScopes - see [how they are interpreted on the other end](https://github.com/taskcluster/taskcluster-proxy/blob/9bb2082b1098b3e9c0e45910e6ceeef96016e10a/main.go#L143-L145).

tcproxy/tcproxy_test.go
djmitche: It duplicates another test name, and caused quite a bit of confusion trying to figure out which one had failed.  I can use another name if you'd prefer, as long as it's not a duplicate.

taskcluster_proxy_test.go
djmitche: I'll add the comments. `curlget.go` is only used by this test, so I don't think it warrants generalization.

djmitche:
> However, after this PR, `env` is `[]string{"TASKCLUSTER_PROXY_URL=.."}`, and that is put into `cmd.Env` unchanged, so all of the other environment variables disappear.

This was slightly off-base: `env` is *not* `[]string{"TASKCLUSTER_PROXY_URL=.."}`, it is still `nil`.  But the taskcluster_proxy feature later appends `"TASKCLUSTER_PROXY_URL=.."` to cmd.Env, changing that field from `nil` to `[]string{"TASKCLUSTER_PROXY_URL=.."}`.

I added some debugging to catch the case where the process environment is required and accessToken is nil, but that doesn't catch this particular error.

So, I think editing `cmd.Env` in a feature's setup function is not a good idea.  The options, then, are

 * Change the order of feature initialization and put the value in Payload.Env
 * Special-case this feature and add a `set` command to the batch script

I'm leaning toward the second one.

djmitche:
It's green!

process/process_windows.go
```diff
@@ -59,6 +59,15 @@ func NewCommand(commandLine []string, workingDirectory string, env []string, acc
 			return nil, err
 		}
 		env = *environment
+	} else {
+		// if env is nil, then cmd.Env will be nil which meansa to copy the
+		// parent process's environment, which is exactly what we want.  But if
+		// env is not nil, then the intent was to extend the current list of
+		// environment variables with those in `env`, and without an
+		// accessToken that's not possible.
```
petemoore: This is close, but not quite right. See [the process handling source code](https://github.com/golang/go/blob/go1.11.4/src/os/exec_posix.go#L39-L41) from the go standard library (go 1.11.4). It does as you say; that is, if `env` is nil, env is set to `os.Environ()`. Instead we can write this whole block like this:

```go
    var err error
    var environment *[]string
    if accessToken == 0 { 
        // Running as current user; merge env with current environment.
        environment, err = win32.MergeEnvLists(&os.Environ(), &env)
    } else {
        // Running as task user; create environment block and merge env with it.
        environment, err = win32.CreateEnvironment(&env, accessToken)
    }   
    if err != nil {
        return nil, err 
    }   
    env = *environment
```

plat_all-unix-style.go
```diff
@@ -121,9 +121,18 @@ func (task *TaskRun) EnvVars() []string {
 		taskEnv[k] = v
 	}
 	taskEnv["TASK_ID"] = task.TaskID
+	taskEnv["TASKCLUSTER_ROOT_URL"] = config.RootURL
+
+	// see if we should include TASKCLUSTER_PROXY_URL
+	tcProxyFeature := &TaskclusterProxyFeature{}
+	if tcProxyFeature.IsEnabled(task) {
+		taskEnv["TASKCLUSTER_PROXY_URL"] = tcProxyFeature.getTaskclusterProxyURL()
+	}
```
petemoore: nit: This should really be done inside the feature (i.e. inside `taskcluster_proxy.go`). Extra points for implementing a generic method to augment an environment, which uses `win32.MergeEnvLists` on Windows, or just appends to the env for non-windows (like you had before). This could be in `plat_windows.go` and `plat_all-unix-style.go`, for example.

Note, with the suggested fix from my previous comment, `env` is always expanded to the full environment (i.e. not nil) so augmenting it suffices (no need to consider case of `env == nil`).

If you don't feel inclined to do this, I'm happy to implement it in a follow-up PR which I can send to you for review. Your version works correctly, it just has the architectural disadvantage that the core now has taskcluster proxy feature concerns.

tcproxy/tcproxy_test.go
petemoore: That's a very good reason, thank you!

plat_all-unix-style.go
djmitche: I did this.  I like the design a good bit more.  I did it in Start instead of NewTaskFeature because that method can return an error.

process/process_windows.go
djmitche: Done, thanks for the sample code.

djmitche:
OK, that's all green except for coveralls.  I don't know why it thinks coveraged dropped 52.9% - that's obviously wrong!

plat_windows.go
```diff
@@ -396,6 +397,20 @@ func (task *TaskRun) prepareCommand(index int) *CommandExecutionError {
 	return nil
 }
 
+// Set an environment variable in each command.  This can be called from a feature's
+// NewTaskFeature method to set variables for the task.
+func (task *TaskRun) setVariable(variable string, value string) error {
```
petemoore: This is just a nit, not a required change.

Nit: I would have probably called this `setEnvironmentVariable` for clarity, and made it a function of _the environment_ rather than the task, but it doesn't really matter. This avoids the duplication of the loop over the task commands in both platform implementations. If you provide a method that just updates an environment block in a platform-independent way, you can have a single loop over the task commands, that is platform independent. It doesn't really matter though.

Note, originally the `windows` and `all-unix-style` process implementations were quite different, but with go 1.10 it became possible to use the standard library for running processes as a different user under windows, at which point it became possible to have very similar implementations for both. I intend to clean up some of the duplication in a future PR, and at that point will probably introduce a separate go type for an environment block (something like `type Environment []string`) that will have its own `func (env *Environment) Set(envVar string, value string)`, `func (env *Environment) Clear(envVar string)` functions for modifying it. So it makes sense to rework this at the same time, so we can leave it as is for now.

djmitche: I'll leave that to you :)

PR #137, opened 2019-01-16
bug 1518913 - add ed25519 support

### LINKED ISSUE #136: bug 1518913 - add ed25519 cot signature support; deprecate gpg
GPG 2.0.x is past its EOL; our current cot-gpg-keys solution is high maintenance and can only be properly tested on puppetized scriptworkers. Moving to a more modern signature algorithm and a set of known public keys, without a web of trust, will improve both of these situations.

We should:

- add ed25519 cot signature support to generic-worker.
- deprecate gpg support; once all 3 worker implementations are uploading signed ed25519 cot artifacts, we'll drop gpg support across the board.
- leave `chainOfTrust.json.asc` alone, until we drop gpg support
- create and upload two new artifacts: an unsigned `chain-of-trust.json`, and an ed25519 signature `chain-of-trust.json.sig`.

Ideally, I'd like to get the solutions in all 3 worker implementations working before we roll out, to avoid churn. I'm signing up to write this patch, though I may need a hand with both generic-worker and golang questions.

See also: https://github.com/mozilla-releng/scriptworker/pull/294 and the discussion in https://github.com/mozilla-releng/scriptworker/pull/293#issuecomment-451339959 .

escapewindow:
Also renames the current cot variables to add `openpgp`, for easier removal later.

This is very much a work in progress. @petemoore, non-urgent, but if you have pointers or if I'm doing anything clearly wrong, I'd love to hear. I'm renaming a bunch of variables and functions, which may or may not be what we want.

Fixes #136.


escapewindow:
I *think* this is in decent shape at this point. I'd love pointers on how to test further.

all-unix-style.yml
```diff
@@ -135,10 +135,10 @@ properties:
         type: boolean
         title: Enable generation of a openpgp signed Chain of Trust artifact
```
tomprince: This line still mentions just openpg. I suspect just `s/openpgp//` is the appropriate fix.

docs/features.md
```diff
@@ -29,22 +29,24 @@ Example:
 
 #### Since: generic-worker 5.3.0
```
tomprince: This should probably mention the version that ed25519 support was added somewhere.

escapewindow: Yeah, I wasn't sure how best to do that. Added another `#### Since:` line with a `TBD`; maybe that will work. We could add another `#### Since:` line when we remove gpg support.

all-unix-style.yml
escapewindow: Done.

escapewindow:
@petemoore is the -53.5% decrease [!] in code coverage a concern?

Looking at the windows bustage, not sure how far I'll get.

escapewindow:
@petemoore this is ready for review. I would prefer to finish adding ed25519 support to all 3 worker implementations and test them together before we finalize, but this looks like this should work with https://github.com/mozilla-releng/scriptworker/pull/294 .

(I imagine this will also need some generic-worker deployment changes for the new keys.)

chain_of_trust_all-unix-style.go
```diff
@@ -12,13 +12,19 @@ import (
...
+		config.OpenPGPSigningKeyLocation,
+		config.Ed25519SigningKeyLocation,
 	}
-	r := c.Execute()
-	if !r.Failed() {
-		return fmt.Errorf(ChainOfTrustKeyNotSecureMessage)
+	for _, path := range signingKeyPaths {
+		c, err := process.NewCommand([]string{"/bin/cat", path}, cwd, cot.task.EnvVars())
+		if err != nil {
+			panic(fmt.Errorf("SERIOUS BUG: Could not create command (not even trying to execute it yet) to cat private chain of trust key - %v", err))
```
petemoore: nit - might be nice to add `path` variable to the panic message to show which one it failed on

chain_of_trust_windows.go
```diff
@@ -12,31 +12,43 @@ import (
...
+		config.Ed25519SigningKeyLocation,
 	}
-	r := c.Execute()
-	if !r.Failed() {
-		log.Print(r.String())
-		return fmt.Errorf(ChainOfTrustKeyNotSecureMessage)
+	for _, path := range signingKeyPaths {
+		c, err := process.NewCommand([]string{"cmd.exe", "/c", "type", path}, cwd, nil, accessToken)
+		if err != nil {
+			panic(fmt.Errorf("SERIOUS BUG: Could not create command (not even trying to execute it yet) to cat private chain of trust key - %v", err))
```
petemoore: here too

docs/features.md
```diff
@@ -29,22 +29,28 @@ Example:
...
 
-Enabling this feature will mean that the generic worker will publish an
-additional task artifact `public/chainOfTrust.json.asc`. This will be a clear
-text openpgp-signed json object, storing the SHA 256 hashes of the task
-artifacts, plus some information about the worker. This is signed by a openpgp
-private key, both generated and stored on the worker. This private key is never
+This feature was added in generic-worker 5.3.0, with gpg support.
+
+#### Since: generic-worker TBD
```
petemoore: My prediction is this will 11.2.0, but we'll only really know at release time.

runasadministrator_windows_test.go
```diff
@@ -85,11 +85,12 @@ func TestRunAsAdministratorMissingOSGroup(t *testing.T) {
 	_ = submitAndAssert(t, td, payload, "exception", "malformed-payload")
 }
 
+// TODO can we parameterize these to test both openpgp + ed25519 paths?
```
petemoore: I think we can replace this `TODO` with a comment that the payload isn't actually executed, because we should always get `malformed-payload` task resolution. Therefore, the `type` command should never actually run, so it doesn't matter too much what it does. The reason I put the `type` command in is in case it really does run, and then we see if it can actually log the private key file - then we know we're really in trouble! 😟

escapewindow:
Thank you!

I changed the `runasadministrator_windows_test.go` tests to point at the ed25519 keys, so it's a smaller patch to tear out openpgp support later.

petemoore:
This looks perfect. The test failure should be fixed by #139 but let's wait for that to get reviewed and landed, and then we can merge here, to make sure it fixes the test failures.

petemoore:
The failures were not serious:

* code coverage reduction in travis, due to PR coming from a forked repo that doesn't have sufficient credentials to run all tests
* timeout in windows 10 tests, probably due to having a slow instance in ec2. At the time it failed, it had passed all tests, and was almost at the end of the test run

Based on this, I'll merge.

PR #143, opened 2019-02-01
Fix typo ('deploment')

djmitche:


petemoore:
Sorry @djmitche I saw this too late, and fixed it in #144 before I noticed you had already created #143.

PR #144, opened 2019-02-01
Bug 1524630 - `generic-worker --help` improvements

petemoore:
`generic-worker --help` should state that `*BaseURL` settings are optional overrides for the global `rootURL` setting.

See [bug 1524630](https://bugzil.la/1524630) for context.

Note, the `README.md` update is just a consequence of running the `refresh_readme.sh` script in the root directory, which regenerates `README.md` based on the output of `generic-worker --help`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ and reports back results to the queue.
 
           accessToken                       Taskcluster access token used by generic worker
                                             to talk to taskcluster queue.
-          clientId                          Taskcluster client id used by generic worker to
+          clientId                          Taskcluster client ID used by generic worker to
                                             talk to taskcluster queue.
           ed25519SigningKeyLocation         The ed25519 signing key for signing artifacts with.
           livelogSecret                     This should match the secret used by the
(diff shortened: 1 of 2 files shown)


petemoore:
@djmitche the coveralls failure is just due to a -0.3% decrease in coverage due to a couple of extra doc lines - so not a blocker! 😆

README.md
```diff
@@ -192,17 +192,22 @@ and reports back results to the queue.
                                             for serving live logs; see
                                             https://github.com/taskcluster/livelog and
                                             https://github.com/taskcluster/stateless-dns-server
-          rootURL                           The root URL of the Taskcluster deploment to which
+          rootURL                           The root URL of the taskcluster deployment to which
                                             clientId and accessToken grant access. For example,
-                                            'https://taskcluster.net'.
+                                            'https://taskcluster.net'. Individual services can
+                                            override this settings - see the *BaseURL settings.
```
djmitche: *this setting (singular)

README.md
```diff
@@ -192,17 +192,22 @@ and reports back results to the queue.
...
 
         ** OPTIONAL ** properties
         =========================
 
-          authBaseURL                       The base URL for API calls to the auth service.
+          authBaseURL                       The base URL for taskcluster auth API calls.
+                                            If not provided, the base URL for API calls is
+                                            instead derived from rootURL setting as follows:
+                                              * https://auth.taskcluster.net/v1 for rootURL https://taskcluster.net
+                                              * <rootURL>/api/auth/v1 for all other rootURLs
```
djmitche: It seems like you could just leave this at "derived from the rootURL setting".  Within six months we'll be removing the taskcluster.net special-case.

README.md
petemoore: Nice spot!

README.md
petemoore: I think having the extra clarity is useful, and we can still remove it in 6 months if the special casing goes away.

djmitche: It's not an "if" :)

PR #146, opened 2019-02-04
Bug 1375200 - fetch secrets from taskcluster secrets service

petemoore:
This is a bit of an overhaul of the secrets management of the worker.

Now "public" generic worker config should be stored in the `userData` property of a worker type definition (rather than `secrets`), and "private" generic worker config should be stored in a secret called `worker-type:<provisionerId>/<workerType>` in the taskcluster-secrets service.

I've partitioned the config properties into public and private sets, and enforced that private config properties come from the taskcluster secrets service, and public properties come from the worker type definition. Furthermore I've made it a fatal error if config settings in the worker type definition `userData` property are not known public generic-worker config properties, or if the secrets config properties are not private ones. This may seem a bit heavy-handed, but I wanted to avoid that config files are copied/pasted into taskcluster-secrets without concern for whether the properties should be visible in the worker type definition or not. This makes things pretty strict, but helps to ensure that private properties stay private, and public ones stay public. Furthermore any unrecognised properties will also cause a fatal error - so this helps avoid problems of misspelled config properties entering worker type definitions or secrets settings.

I'm creating this PR now, but will assign reviewers as soon as I'm happy with it, all the CI tasks are passing, etc.

Note, when this lands, it will be a major version bump as this is backwardly incompatible behaviour. We will also need to update worker type definitions prior to rolling out the new AMIs. However, the new worker type definitions can include the new `userData` field without breaking the old workers, and the old `secrets` field will not break the new workers, so the worker type changes can be landed before rolling out the new workers, and then the secrets can be removed from the worker type definitions after the new workers are running successfully.

See [bug 1375200](https://bugzil.la/1375200).


petemoore:
@jhford, note the only failure above is that code coverage has dropped by 0.07%. Everything else in the CI passed. I think the 0.07% drop is acceptable.

PR #149, opened 2019-02-08
Bug 1375200 - extend test coverage and apply fixes

petemoore:
Hi John!

I added a bunch more tests and caught a couple more conditions, can you take a look again? I think this should be the last PR for this bug! :-)

Note the diff is a bit ugly since I rewrote the tests - probably easiest to look at [aws_test.go](https://github.com/taskcluster/generic-worker/blob/bug1375200_part4/aws_test.go) directly, rather than the diff of the changes to it.

If you spot any test cases I've not considered, let me know. This is my biggest area of concern. Many thanks John!


jhford:
Oh and as far as possibly missing tests, the only test which I can think of is to ensure that every single time that the remove credential endpoint is called.  Not calling that endpoint is an absolutely critical bug.  Maybe that's already asserted in the tests and I don't see it?

petemoore:
> Oh and as far as possibly missing tests, the only test which I can think of is to ensure that every single time that the remove credential endpoint is called. Not calling that endpoint is an absolutely critical bug. Maybe that's already asserted in the tests and I don't see it?

That's a very good idea - I'll add that! Thanks!

jhford:
Nice approach to that test, r+ (can't reflag it) just to avoid ambiguity

PR #150, opened 2019-02-19
add additional exit code

aerickson:
I was debugging why generic-worker wouldn't start and found that the same error code was returned for two states. This PR attempts to make debugging easier.


petemoore:
Many thanks!

PR #152, opened 2019-02-26
Simplify DeepSpeech Windows worker setup

lissyx:


lissyx:
Looks like this will need more thinking.

PR #156, opened 2019-03-04
Bug 1436195 - implement exposing services publicly

djmitche:
This introduces a new package designed to expose services available on
the worker's loopback interface to the public Internet.  The services
can include both HTTP services, which are proxied directly, and TCP
services, which are made available via binary messages on a websocket.

### DIFF
--- a/expose/local/local_test.go
+++ b/expose/local/local_test.go
@@ -1,0 +1,71 @@
+package local
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/taskcluster/generic-worker/expose"
+)
+
+func TestConstructor(t *testing.T) {
+	exposer, err := New(net.ParseIP("127.0.0.1"))
+	if err != nil {
+		t.Fatalf("Constructor returned an error: %v", err)
+	}
+	if exposer == nil {
+		t.Fatalf("Constructor did not return a value")
+	}
+}
+
+func makeExposer(t *testing.T) expose.Exposer {
+	exposer, err := New(net.ParseIP("127.0.0.1"))
+	if err != nil {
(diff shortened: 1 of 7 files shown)


djmitche:
Pete, could you give this a quick look before I proceed?  I don't want to go too far if this is a totally bogus idea.

My plans from here are:

 * implement an Exposer interface that uses statelessdns
 * modify livelog to use Exposer
 * implement an Exposer interface that uses websocktunnel

This leaves ample opportunity to proxy RDP over WS, and/or support the docker-worker mode of proxying VNC and terminal over WS, when those features are implemented.

I've written automated tests for the important bits here, but also validated this using a bunch of short test programs and various http, websocket and tcp clients and servers.

expose/local/local_test.go
```diff
@@ -0,0 +1,71 @@
...
+	exposer := makeExposer(t)
+	exposure, err := exposer.ExposeHTTP(uint16(testPort))
+	if err != nil {
+		t.Fatalf("ExposeHTTP returned an error: %v", err)
+	}
+	defer exposure.Close()
+
+	gotURL := exposure.GetURL()
+	host, port, _ := net.SplitHostPort(gotURL.Host)
+	assert.Equal(t, "http", gotURL.Scheme, "Should return URL with correct schem")
```
petemoore: scheme

expose/local/local_test.go
```diff
@@ -0,0 +1,71 @@
...
+	res, err := http.Get(gotURL.String())
+	if err != nil {
+		t.Fatal(err)
+	}
+	greeting, err := ioutil.ReadAll(res.Body)
+	res.Body.Close()
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.Equal(t, "Hello, world", string(greeting), "got greeting via proxy")
```
petemoore: Let's also assert a 200 status code. Perhaps a non-200 status code test would also be good, both with and without a response body.

djmitche: I'll add that to proxy_test.go.

djmitche:
I'll open a new PR when the next few steps are done.

PR #159, opened 2019-03-05
Bug 1532588 - use reflect.DeepEqual rather than fmt.Sprintf("%q") for testing equality

petemoore:
See [bug 1532588](https://bugzil.la/1532588) for details.


petemoore:
Note, the pending win7/win10 CI tasks are due to [an OCC deployment issue](https://github.com/mozilla-releng/OpenCloudConfig/commits/master) which Rob has kindly offered to take a look into tomorrow. Those worker types are currently broken.

PR #160, opened 2019-03-05
deepspeech-win: Add missing .Net Framework infra

lissyx:


lissyx:
@petemoore **hopefully** this should be good, I cannot reproduce any build error locally (I should have checked more carefully :/).

PR #163, opened 2019-03-07
add a 'dep check' run to CI

djmitche:


djmitche:
I also filed https://github.com/getsentry/raven-go/issues/243

.taskcluster.yml
```diff
@@ -494,3 +494,49 @@ tasks:
...
+      - secrets:get:repo:github.com/taskcluster/generic-worker
+      - docker-worker:cache:generic-worker-checkout
+    payload:
+      features:
+        taskclusterProxy: true
+      maxRunTime: 3600
+      image: golang:1.10
+      env:
+        # once bug 1508383 is landed, this won't be required (the worker will supply it)
+        TASKCLUSTER_ROOT_URL: https://taskcluster.net
```
petemoore: I think this isn't needed as bug 1508383 has landed, but I'll remove it in a separate PR as we have that throughout the `.taskcluster.yml` file.

djmitche: It is still needed for generic-worker workers that have not been upgraded to ^13.0.0

PR #168, opened 2019-03-11
Bug 1534223 - first stab

petemoore:
Use taskcluster-proxy in CI integration tests.


petemoore:
I've put this PR on ice, in light of bug 1548108 which is blocking it. Hopefully it won't bitrot too much, until I have some time to return to it.

petemoore:
Despite the bitrot, I'd like to return to this at some point, just isn't a priority quite at the moment.

PR #169, opened 2019-03-12
deepspeech-win-gpu: Create with CUDA + NVIDIA Driver

lissyx:


lissyx:
Hold on @petemoore I just noticed we will soon need CUDA v10

lissyx:
@petemoore it should be good now :)

PR #175, opened 2019-03-26
Bug 1502335 - Support running a task inside a fixed docker image

petemoore:
See [bug 1502335](https://bugzil.la/1502335) for more context.


petemoore:
Hey Brian!

This is essentially a no-op for all platforms other than docker-linux. I've intentionally disabled tests that require [features which are not yet implemented](https://bugzilla.mozilla.org/showdependencytree.cgi?id=1499055&hide_resolved=1).

The choice of native engine or docker engine is made at build time via a build constraint (`nativeEngine` or `dockerEngine`). The majority of the code changes here is the addition of the docker client and its dependencies, which are vendored in this repository.

I pulled the tests that require not-yet-implemented features into separate files, and put them behind a `!dockerEngine` build constraint. As the remaining features get added in subsequent bugs, the intention is to move them back, until all tests are supported with the docker engine, and we have no more `!dockerEngine` build constraints.

Gopkg.toml
```diff
@@ -25,6 +25,10 @@
 #   unused-packages = true
 
 
+[[constraint]]
```
imbstack: Oh neat. TIL about how this all works.

petemoore:
Thanks Brian!

PR #176, opened 2019-03-27
Bug 1436195 - add websocktunnel exposer

djmitche:
This should wrap up the changes needed to generic-worker.

The bulk of this patch is adding a WST-based exposer, but that exposes precisely the same interface as the other exposers, so the last commit -- using the WST exposer when given appropriate configuration -- is pretty simple.  Is there anything else I need to do to add new configuration options like this?

The HTTP proxying and copying going on for some of these connections is kind of mind-meltingly complex, but being composed of simple pieces it all seems to work.  I think the most complex is for a WST-exposed TCP port:
tunnels

Once this lands, I think the next step will be to run it on a test workerType and see it at work..


gwconfig/config.go
```diff
@@ -68,6 +68,8 @@ type (
 		WorkerID                       string                 `json:"workerId"`
 		WorkerType                     string                 `json:"workerType"`
 		WorkerTypeMetadata             map[string]interface{} `json:"workerTypeMetadata"`
+		WstAudience                    string                 `json:"wstAudience"`
+		WstServerURL                   string                 `json:"wstServerURL"`
```
petemoore: Please update the usage docs (`generic-worker --help`) with details about config settings `wstAudience` and `wstServerURL`. Note that required properties and optional properties are listed in separate sections of the usage docs.

nit: the Go naming convention prefers uppercase for acronyms, i.e. `WSTAudience` / `WSTServerURL`.

djmitche:
Updated to the latest WST without the workaround for https://github.com/taskcluster/websocktunnel/issues/41

djmitche:
Updated to the latest WST without the workaround for https://github.com/taskcluster/websocktunnel/issues/41.  WST spelling fixed, and docs added.

PR #177, opened 2019-03-28
bug 1534506 - Remove cot openpgp usage.

escapewindow:
We can probably also remove the vendored openpgp module. I wasn't sure if I should do that here?

This can roll out any time after scriptworker 23.0.0 rolls out. (puppet patch [here](https://github.com/mozilla-releng/build-puppet/pull/439).)


escapewindow:
I tried unvendoring openpgp, and got a dep test failure. I'll leave that for now.

escapewindow:
Looks better :)

escapewindow:
@grenade : we can remove the cot openpgp references from occ when this lands. The current gpg keypair is expiring within 60 days, but nothing is looking at the signatures anymore, so we probably want to just kill the keypair, rather than rotate it.

PR #184, opened 2019-04-04
deepspeech-win: Do DisableMSI at the begining

lissyx:


worker_types/deepspeech-win/userdata
```diff
@@ -30,6 +30,17 @@ Set-Service "WinDefend" -StartupType Disabled -Status Stopped
 # Disable disk indexing
 Set-Service "WSearch" -StartupType Disabled -Status Stopped
 
+# Toggle DisableMSI to allow Python MSI packages installation with pyenv-win
```
petemoore: I think you should add a comment here that you are doing this as the very first step, because when you ran it after granting `SeCreateSymbolicLinkPrivilege` to `Everyone` group, that permission was deleted.

PR #185, opened 2019-04-04
Bug 1533694 - prepare next task user before using current task user in case of unintended reboot

petemoore:
See [bug 1533694](https://bugzil.la/1533694) for context.

I slipped in a couple of refactor/cleanups at the same time, which unfortunately added a little noise to the diff.


plat_all-unix-style.go
```diff
@@ -187,22 +181,27 @@ func defaultTasksDir() string {
...
+				return err
+			}
+		}
+	}
+	return nil
 }
 
-func deleteTaskDirs() {
-	removeTaskDirs(config.TasksDir)
+func GrantSIDFullControlOfInteractiveWindowsStationAndDesktop(sid string) (err error) {
```
imbstack: I love this function name

runtime/runtime_darwin.go
```diff
@@ -0,0 +1,35 @@
+package runtime
+
+import (
+	"log"
+	"os/exec"
+)
+
+func (user *OSUser) Create(okIfExists bool) error {
```
imbstack: `ofIfExists` seems unused here. I'm guessing that's intentional, just wanted to make sure.

usage.go
```diff
@@ -0,0 +1,315 @@
+package main
+
+import (
+	"runtime"
+	"strconv"
+)
+
+func usage(versionName string) string {
```
imbstack: Moving this out of `main` is great.

runtime/runtime_darwin.go
petemoore: Ah, I wanted to harmonise the function call across Windows and other platforms, so it could be called from platform-agnostic code. Currently user creation on darwin and linux is a TODO, so this code is currently uncalled.

plat_all-unix-style.go
petemoore: :-)

usage.go
petemoore: Thanks!

PR #166, opened 2019-03-08
Bug 1436195 - implement exposer, use it for livelog

djmitche:
I'd like to get this much merged soon.  It does not yet support websocktunnel (there's a blocking bug for that), but is most of the groundwork to support that once it's ready.

The last commit is a little "iffy": it gives a static lifetime for each exposure, rather than setting the lifetime to the maxRunTime of a task.  However, this only applies for stateless-dns, not websocktunnel, so it's a temporary measure and, I think, not too bad.

I ran `build.sh -t` successfully, but it never actually failed on my changes to `/livelog.go` -- so I wonder if that file is well-tested.  Advice for ways to give this additional testing are appreciated.  The tests for `./livelog/...` and `./expose/...` seem more thorough, and do pass.

PR #189, opened 2019-04-10
Reland exposer

djmitche:
This relands #166 and #176.  Tests on master aren't passing for me locally, so let's see how they do in CI..


djmitche:
The windows test failed for what looks like an unrelated reason.  I've re-run it.

djmitche:
All green in TC :)

djmitche:
BTW, I ran the linux tests in a loop locally and then forgot about it, so it ran some many thousands of time without error.

livelog.go
```diff
@@ -153,30 +157,34 @@ func (l *LiveLogTask) reinstateBackingLog() {
...
+	exposeURL := l.exposure.GetURL()
+	if exposeURL.Path == "/" {
+		exposeURL.Path = logURL.Path
 	} else {
-		getURL.Scheme = "http"
+		exposeURL.Path = exposeURL.Path + logURL.Path
 	}
-	getURL.Host = statelessHostname + ":" + strconv.Itoa(int(l.liveLog.GETPort))
+
+	expires := time.Now().Add(time.Duration(l.task.Payload.MaxRunTime+900) * time.Second)
```
petemoore: a comment to explain that the 900 is to account for clock drift would be useful here

livelog/livelog.go
```diff
@@ -21,14 +21,11 @@ import (
 // LiveLog provides access to a livelog process running on the OS. Use
 // New(liveLogExecutable string) to start a new livelog instance.
 type LiveLog struct {
-	SSLCert string
-	SSLKey  string
-	secret  string
-	PUTPort uint16
-	GETPort uint16
-	putURL  string
-	// The fully qualified HTTP GET URL where the log will be published.
```
petemoore: Did you intend to delete this comment for `GetURL`?

PR #191, opened 2019-04-11
Bug 1436195 - fix livelog certs, port

djmitche:
The first fixes a silly bug: I was passing filenames where PEM data was expected.

The second is a little more significant: without this change, exposer serves exposed things on an arbitrary port, but of course we don't have arbitrary ports open.  We have a single port open, which is labeled "livelogGETPort" in the configuration.  So, we configure the stateless DNS exposer to always expose on that port.  The similarly-named "livelogPUTPort" is not really similar -- it's used for communication between the task execution and the livelog process.  That's really an implementation detail and should probably be moved out of the config eventually.  And, once stateless DNS is turned off, we can get rid of `livelogCertificate`, `livelogGETPort`, `livelogKey`, and `livelogSecret`.

This ran a task with logs served over stateless DNS.  See it in action at https://tools.taskcluster.net/groups/aYhpWoTPSAWm47yk_VW2vg/tasks/aYhpWoTPSAWm47yk_VW2vg (or more accurately, by creating a new task like that one).

### DIFF
--- a/expose/statelessdns.go
+++ b/expose/statelessdns.go
@@ -13,6 +13,7 @@ import (
 
 type statelessDNSExposer struct {
 	publicIP           net.IP
+	publicPort         uint16
 	hostDomain         string
 	statelessDNSSecret string
 	duration           time.Duration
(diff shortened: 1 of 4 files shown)


petemoore:
Awesome, many thanks Dustin!

PR #194, opened 2019-04-24
look for tooltool in the new location

djmitche:
This recently moved -- https://github.com/mozilla/build-tooltool/

### DIFF
--- a/worker_types/ami-test-win2012r2/userdata
+++ b/worker_types/ami-test-win2012r2/userdata
@@ -77,7 +77,7 @@ $acl.AddAccessRule($ace)
 Set-Acl "C:\builds" $acl
 
 # download tooltool
-$client.DownloadFile("https://raw.githubusercontent.com/mozilla/build-tooltool/master/tooltool.py", "C:\builds\tooltool.py")
+$client.DownloadFile("https://raw.githubusercontent.com/mozilla/release-services/master/src/tooltool/client/tooltool.py", "C:\builds\tooltool.py")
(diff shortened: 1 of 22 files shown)


djmitche:
Oh, and worth noting, that this built an image without complaining that the URL download failed, leading to https://tools.taskcluster.net/groups/fmtw81BoQsGNPQG3XCvWPw/tasks/fmtw81BoQsGNPQG3XCvWPw/runs/0/logs/public%2Flogs%2Flive.log -- it would be great if image building failed when a download failed!

petemoore:
> Oh, and worth noting, that this built an image without complaining that the URL download failed, leading to https://tools.taskcluster.net/groups/fmtw81BoQsGNPQG3XCvWPw/tasks/fmtw81BoQsGNPQG3XCvWPw/runs/0/logs/public%2Flogs%2Flive.log -- it would be great if image building failed when a download failed!

I think the best way for us to handle this will be in [RFC 122](https://github.com/taskcluster/taskcluster-rfcs/issues/122). With the current machine image creation mechanics, there is no easy way to signal that a shutdown instance shouldn't be used, since the worker_type.sh script simply waits for the instance to be in shutdown state before snapshotting it to create the machine image. There isn't something like an exit code that it can check first. However with RFC 122 the image creation will be a taskcluster task itself, which will have a resolution (successful, failed, exception) and we'll have a log of the image setup as a taskcluster artifact too, so failure-handling will be much simpler.

PR #195, opened 2019-04-26
Fix test TestRemoveTaskDirs

petemoore:
Looks like I inadvertently broke this test when I landed [bug 1533694](https://bugzil.la/1533694) in release 14.0.2.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1201,3 +1201,27 @@ type TaskContext struct {
 	TaskDir string
 	User    *gwruntime.OSUser
 }
+
+// deleteTaskDirs deletes all task directories (directories whose name starts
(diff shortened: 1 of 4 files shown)


petemoore:
Note: I reverted the commit in this PR due to [this bug](https://sentry.prod.mozaws.net/operations/generic-worker/issues/5695850/) in [this commit](https://github.com/taskcluster/generic-worker/commit/515a8eba4908a6ed269f7f836dc2e8d272ad370a) and then pushed a corrected version in [this commit](https://github.com/taskcluster/generic-worker/commit/f444c25128b8bd34a5c9257fd97e69444b575308).

PR #198, opened 2019-05-01
README.md: mention build.sh in building from source instructions

milescrabill:
it wasn't immediately clear to me why `go get github.com/taskcluster/generic-worker` resulted in errors, this should clear some of that up

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ If you prefer not to use a prepackaged binary, or want to have the latest unrele
 * Run `go get github.com/taskcluster/livelog`
 * Run `go get github.com/taskcluster/taskcluster-proxy`
 
+Note that the build process requires custom tooling, and your `go get` may mention errors.
+
+Run `./build.sh` to build binaries, ensure dependencies are vendored, and generate code.
+
+`./build.sh` takes optional arguments, `-a` to build all platforms, and `-t` to run tests. By default tests are not run and only the current platform is built.
+
 All being well, the binaries will be built under `${GOPATH}/bin`.
 
 # Acquire taskcluster credentials for running tests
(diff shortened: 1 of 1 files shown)


milescrabill:
I'm not sure what's going on with Coveralls here, but that doesn't look right.

README.md
```diff
@@ -26,6 +26,12 @@ If you prefer not to use a prepackaged binary, or want to have the latest unrele
 * Run `go get github.com/taskcluster/livelog`
 * Run `go get github.com/taskcluster/taskcluster-proxy`
 
+Note that the build process requires custom tooling, and your `go get` may mention errors.
```
petemoore: Nice spot! Indeed that was broken, but luckily we can fix it, so I would prefer to replace your line 29 with the following change to line 25:

* Run `go get -tags nativeEngine github.com/taskcluster/generic-worker` (all platforms) and/or `go get -tags dockerEngine github.com/taskcluster/generic-worker` (linux only)

The reason for this is, for the first build, this is entirely sufficient, since no code changes have been made, so assuming CI has passed on the master branch, the additional checks that build.sh makes should have already been made (e.g. code has been generated, tests compile, etc etc).

README.md
```diff
@@ -26,6 +26,12 @@ If you prefer not to use a prepackaged binary, or want to have the latest unrele
...
+
+Run `./build.sh` to build binaries, ensure dependencies are vendored, and generate code.
```
petemoore: I would prefer: 

> Run `./build.sh` to check go version, generate code, build binaries, compile (but not run) tests, perform linting, and ensure there are no ineffective assignments in go code.

Note: currently `build.sh` does not check dependencies are vendored, but we could add `dep check` to `build.sh` so that it does, and then keep the "ensure dependencies are vendored" part of your edit, or we could just remove it. _At the moment_ only the CI performs checks dependencies are suitably vendored [here](https://github.com/taskcluster/generic-worker/blob/v14.1.0/.taskcluster.yml#L613), rather than `build.sh` but it would probably be nice to have it in `build.sh` too, as it runs pretty quickly so won't steal a lot of time from the build process.

README.md
milescrabill: Fixed!

PR #203, opened 2019-05-14
Bug 1551164 - move create/purge task directory steps into task-claim-and-execute loop

petemoore:
For workers that run more than one task per generic-worker invocation, the task directory creation and purging logic was broken with the code changes from [bug 1533694](https://bugzil.la/1533694), meaning that only a single task directory was created per generic-worker invocation, rather than between every task.

This impacted macOS and linux workers. The fix is to move the task directory creation and task directory purging into the main loop that claims and executes tasks.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -115,6 +115,11 @@ tasks:
       - generic-worker:cache:generic-worker-checkout
     payload:
       maxRunTime: 3600
+      env:
+        # Python 2 is needed for TestDesktopResizeAndMovePointer
+        # Powershell is needed for TestMounts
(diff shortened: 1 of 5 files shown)


petemoore:
CC @jdm @SimonSapin @lissyx

petemoore:
It looks like the broken win7 test is due to [bug 1545339](https://bugzilla.mozilla.org/show_bug.cgi?id=1545339#c9). I'll adjust the test to explicitly use python 2 (I think it is using python 3 in the CI).

petemoore:
The single test failure seems unrelated to the code change, and looks like a rare intermittent issue (I've not seen it before).

PR #205, opened 2019-05-17
Bug 1552465 - only prepare new task environment when task has been claimed

petemoore:
See [bug 1552465](https://bugzil.la/1552465) for details.


petemoore:
> Is there a way to avoid the code duplication? I know `main.go` is a humongous bunch of code already, but it seems like this introduces the risk that these two copies of this code will diverge..

We probably could split `main.go` up a little, I think I'll attempt that in a different PR. I've pushed an updated version that removes the code duplication.

PR #209, opened 2019-06-28
Bug 1382668 - Add information to docs about automatically set environment variables

petemoore:
This updates the task payload schemas to explicitly mention which environment variables are automatically set in task commands.

See [bug 1382668](https://bugzil.la/1382668) for context.

### DIFF
--- a/docker_posix.yml
+++ b/docker_posix.yml
@@ -37,6 +37,14 @@ properties:
       }
       ```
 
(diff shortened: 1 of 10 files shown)


djmitche:
..and it might be worth adding `TASKCLUSTER_USER_CREDENTIALS` to that document.

PR #210, opened 2019-07-02
WIP: Implement 1459203 (remove NSSM), 1436274 (use eventlog)

milescrabill:
There's a ton here. Too much!

- adds dependency on `gopkg.in/natefinch/lumberjack.v2` for logfile management
- removes NSSM and modifies docs, userdata in `worker_types`, etc. as such
- changes the way the main `RunWorker` function is called for flexibility, it now accepts a signal chan
- split out config handling and exit code handling into functions
- adds debug logic for panic scenario exits including writing out a crash log in scenarios where other logging might be failing

- support for installing/removing the generic worker service using native go
- support for specifying configuration location at install time
- support for windows event log when run as a service
- logging to logfile by default in windows
- remove need for a bat script wrapper to change the service working directory, cwd is set to parent dir of worker exe

### Usage testing:

- verify: starting the service in interactive mode [x]
- verify: starting the service as a service works properly [x]
- verify: under expected use, can ctrl-c to properly exit from interactive mode []
  - caveat: not running as LocalUser sticks us into a loop that cannot be broken by interrupts!

### TODO:

- fix CI
- actually write tests
- niceify commits
- test: install the service, verify the service exists, verify the event log source exists, remove the service
- test: starting the service without the event log source in non-interactive mode should fail with exit code 78
- test: adding a broken `io.Writer` to the `logWriter` `MultiWriter` will be caught (worth testing?)
- should logfile location be configurable? (currently: generic-worker.log, rotated by `lumberjack`)
- should logfile management be in scope or is that an infra concern?

### Weird things:

- running in interactive mode prints an extra blank newline in between log entries


usage.go
```diff
@@ -7,6 +7,7 @@ const (
 	TASKS_COMPLETE              ExitCode = 0
 	CANT_LOAD_CONFIG            ExitCode = 64
 	CANT_INSTALL_GENERIC_WORKER ExitCode = 65
+	CANT_REMOVE_GENERIC_WORKER  ExitCode = 66
```
petemoore: It might be better to use a different exit code here, as this [used to mean something else](https://github.com/taskcluster/generic-worker/commit/d5a8dd552bed48bd174b43dec40742fc8cfeac74#diff-04c6e90faac2675aa89e2176d2eec7d8L378).

petemoore:
Nice work so far!

milescrabill:
Closing in favor of a branch on this repo for proper testing.

PR #211, opened 2019-07-02
Bug 1499054 - OS task user separation on Linux

petemoore:
Work in progress. See [bug 1499054](https://bugzil.la/1499054) for context.


petemoore:
This work has been moved to #219.

PR #214, opened 2019-07-08
WIP: Implement 1459203 (remove NSSM), 1436274 (use eventlog)

milescrabill:
There's a ton here. Too much!

Continued from https://github.com/taskcluster/generic-worker/pull/210

- adds dependency on `gopkg.in/natefinch/lumberjack.v2` for logfile management
- removes NSSM and modifies docs, userdata in `worker_types`, etc. as such
- changes the way the main `RunWorker` function is called for flexibility, it now accepts a signal chan
- split out config handling and exit code handling into functions
- adds debug logic for panic scenario exits including writing out a crash log in scenarios where other logging might be failing

- support for installing/removing the generic worker service using native go
- support for specifying configuration location at install time
- support for windows event log when run as a service
- logging to logfile by default in windows
- remove need for a bat script wrapper to change the service working directory, cwd is set to parent dir of worker exe

### Usage testing:

- verify: starting the service in interactive mode [x]
- verify: starting the service as a service works properly [x]
- verify: under expected use, can ctrl-c to properly exit from interactive mode []
  - caveat: not running as LocalUser sticks us into a loop that cannot be broken by interrupts!

### TODO:

- fix CI
- actually write tests
- niceify commits
- test: install the service, verify the service exists, verify the event log source exists, remove the service
- test: starting the service without the event log source in non-interactive mode should fail with exit code 78
- test: adding a broken `io.Writer` to the `logWriter` `MultiWriter` will be caught (worth testing?)
- should logfile location be configurable? (currently: generic-worker.log, rotated by `lumberjack`)
- should logfile management be in scope or is that an infra concern?

### Weird things:

- running in interactive mode prints an extra blank newline in between log entries


.gitattributes
```diff
@@ -1 +1,2 @@
 /testdata/** eol=lf
+* text=auto
```
petemoore: I suspect it might be this line that is breaking CI on Windows 10 and Windows Server 2012 R2.

.gitattributes
```diff
@@ -1 +1,2 @@
 /testdata/** eol=lf
+# * text=auto
```
petemoore: Can we just delete this line, or add a comment above it to explain why it is commented out?

README.md
```diff
@@ -351,6 +348,9 @@ and reports back results to the queue.
   Usage:
     generic-worker run                      [--config         CONFIG-FILE]
                                             [--configure-for-aws | --configure-for-gcp]
+    generic-worker install service          [--service-name SERVICE-NAME]
```
petemoore: Can we indent `SERVICE-NAME` a couple of spaces so it lines up with `CONFIG-FILE`?

build.cmd
```diff
@@ -10,7 +10,7 @@ cd gw-codegen
 go install -v || exit /b %ERRORLEVEL%
 cd ..
 go generate || exit /b %ERRORLEVEL%
-go install -v ./... || exit /b %ERRORLEVEL%
+go install -v -tags multiuser ./... || exit /b %ERRORLEVEL%
```
petemoore: Thanks for fixing this!

helper_test.go
```diff
@@ -225,7 +225,8 @@ func execute(t *testing.T, expectedExitCode ExitCode) {
 	if err != nil {
 		t.Fatalf("Test setup failure - could not write to tasks-resolved-count.txt file: %v", err)
 	}
-	exitCode := RunWorker()
+	interruptChan := make(chan os.Signal, 1)
+	exitCode := RunWorker(interruptChan)
```
petemoore: Does this allow us to `Ctrl-C` out of tests now?

helper_windows_test.go
```diff
@@ -116,3 +116,8 @@ func copyTestdataFileTo(src, dest string) []string {
 func singleCommandNoArgs(command string) []string {
 	return []string{command}
 }
+
+func shouldRunAdminTests() bool {
```
petemoore: Might be worth adding a comment here to explain why this is opt-out rather than opt-in (to encourage future developers not to change to opt-in with realising the potential consequences).

multiuser_service_windows.go
```diff
@@ -0,0 +1,296 @@
...
+
+func init() {
+	// default for a service is C:\Windows\system32
+	dir := path.Dir(os.Args[0])
+	err := os.Chdir(dir)
+	if err != nil {
+		exitOnError(INTERNAL_ERROR, err, "Unable to chdir to %q", dir)
+	}
+	log.SetOutput(logWriter)
+	manageLogFile(filepath.Join(path.Dir(os.Args[0]), "generic-worker.log"))
```
petemoore:
```suggestion
	manageLogFile(filepath.Join(filepath.Dir(os.Args[0]), "generic-worker.log"))
```

multiuser_service_windows.go
```diff
@@ -0,0 +1,296 @@
...
+}
+
+func configureService(name, exePath string, args []string) error {
+	config := mgr.Config{
+		DisplayName: name,
+		Description: "A taskcluster worker that runs on all mainstream platforms",
+		// run as LocalSystem because we call WTSQueryUserToken
+		ServiceStartName: "LocalSystem",
+		ServiceType:      windows.SERVICE_WIN32_OWN_PROCESS | windows.SERVICE_INTERACTIVE_PROCESS,
+		StartType:        mgr.StartAutomatic,
```
petemoore: This will cause the installed service to start automatically on boot, but not immediately service when the service is installed, right? If so, that looks fine.

multiuser_service_windows.go
```diff
@@ -0,0 +1,296 @@
...
+	log.Printf("Created service %q with exePath %q, and args %v",
+		name, exePath, args)
+	return nil
+}
+
+func configureEventlogSource(name, exePath string) error {
+	dir := filepath.Dir(exePath)
+
+	// minimal eventlog message file
+	// https://docs.microsoft.com/en-us/windows/desktop/eventlog/message-files
```
petemoore: Thanks for linking to this doc!

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"github.com/taskcluster/slugid-go/slugid"
+)
+
+func unprivilegedRecovery(t *testing.T) {
+	if r := recover(); r != nil {
+		t.Log("A panic can occur if tests are run as an unprivileged user.")
+		t.Log("Disable tests that require administrator privileges by setting `SKIP_ADMINISTRATOR_TESTS` in your environment")
+		t.Logf("Caught panic: %v", r)
```
petemoore:
```suggestion
		t.Fatalf("Caught panic: %v", r)
```

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+	// _, err = m.OpenService(name)
+	// if err == nil {
+	// 	t.Fatalf("Found service %q after it should have been removed", name)
+	// }
+
+	// // verify eventlog is removed
+	// _, err = eventlog.Open(name)
+	// if err == nil {
+	// 	t.Fatalf("Found eventlog source %q after it should have been removed", name)
+	// }
```
petemoore: Don't we need to remove the service here?

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+// 	if exitCode != int(CANT_LOG_PROPERLY) {
+// 		t.Logf("Expected runService() to exit with CANT_LOG_PROPERLY, got: %s", exitError.Error())
+// 		t.Logf("stderr: %s", stderr)
+// 		t.Logf("stdout: %s", stdout)
+// 		t.Fail()
+// 	}
+// } else {
+// 	t.Fatalf("Expected an error from %q: %v", strings.Join(cmd.Args, " "), err)
+// }
+// }
```
petemoore: Can we remove the commented code if things are working now?

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+func TestRunServiceWithBrokenWriter(t *testing.T) {
```
petemoore: Very nice!!

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+	defer unprivilegedRecovery(t)
+
+	if !shouldRunAdminTests() {
+		t.Skipf("SKIP_ADMINISTRATOR_TESTS set, skipping %q", t.Name())
+	}
+
+	// if we are running the test
+	// as opposed to wrapping the test
```
petemoore: I think these comments don't apply now we commented out the test wrapping, right?

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+	logWriter = brokenWriter{}
+
+	// now we try runService in non-interactive mode
+	// which should attempt to use brokenWriter and fail
+	// with CANT_LOG_PROPERLY
+	exitCode := runService(name, false)
+	// as expected, there was an error
+	if exitCode != CANT_LOG_PROPERLY {
+		t.Fatalf("Expected runService() to exit with CANT_LOG_PROPERLY, got: %v", exitCode)
+	}
```
petemoore: Love it!

multiuser_service_windows_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+	select {
+	case <-time.After(time.Second * 5):
+		t.Fatalf("Timeout waiting for status svc.Stopped")
+	case status = <-c:
+		if status.State != svc.Stopped {
+			t.Fatalf("Expected state svc.Stopped, got status %#v", status)
+		}
+	}
```
petemoore: Awesome test!

simple_docker.go
```diff
@@ -105,6 +105,10 @@ func install(arguments map[string]interface{}) (err error) {
 	return nil
 }
 
+func remove(arguments map[string]interface{}) (err error) {
+	return nil
+}
+
```
petemoore: Isn't the `remove` target only needed on Windows?

usage.go
```diff
@@ -17,6 +17,8 @@ const (
 	CANT_CREATE_ED25519_KEYPAIR ExitCode = 75
 	CANT_SAVE_CONFIG            ExitCode = 76
 	CANT_SECURE_CONFIG          ExitCode = 77
+	CANT_LOG_PROPERLY           ExitCode = 78
+	CANT_REMOVE_GENERIC_WORKER  ExitCode = 79
```
petemoore: Please can you add these exit codes to the usage text on exit codes?

usage_windows.go
```diff
@@ -6,18 +6,30 @@ const (
...
+func removeServiceSummary() string {
+	return `
+    generic-worker remove service           [--service-name   SERVICE-NAME]`
+}
+
+func runServiceSummary() string {
+	return `
+    generic-worker run-service              [--service-name   SERVICE-NAME]
+                                            [--config         CONFIG-FILE]
+                                            [--working-directory  DIRECTORY]
```
petemoore: Can we have `SERVICE-NAME`, `CONFIG-FILE` and `DIRECTORY` all lined up?

usage_windows.go
```diff
@@ -32,6 +44,12 @@ func installService() string {
                                             preconditions have been met.`
 }
 
+func removeServiceDescription() string {
+	return `
+    remove service                         This will remove the generic worker
```
petemoore:
```suggestion
    remove service                          This will remove the generic worker
```

usage_windows.go
```diff
@@ -41,11 +59,10 @@ func customTargets() string {
...
 func platformCommandLineParameters() string {
 	return `
-    --nssm NSSM-EXE                         The full path to nssm.exe to use for installing
-                                            the service.
-                                            [default: C:\nssm-2.24\win64\nssm.exe]
     --service-name SERVICE-NAME             The name that the Windows service should be
-                                            installed under. [default: Generic Worker]`
+                                            installed under. [default: Generic Worker]
+    --working-directory DIRECTORY           The working directory the Generic Worker
+                                            service will use. [default: C:\Windows\system32]`
```
petemoore: Since the worker writes files to the working directory, can we change the default to e.g. `C:\generic-worker`? Perhaps the correct drive letter can be determined from the system, in case the operating system isn't installed on C:.

helper_test.go
milescrabill: I believe it does, yes. Were we not able to before?

I changed this to match the signature of the function, which I had changed before running the tests (yikes!) :P

multiuser_service_windows.go
```diff
@@ -0,0 +1,295 @@
...
+	"golang.org/x/sys/windows/svc"
+	"golang.org/x/sys/windows/svc/debug"
+	"golang.org/x/sys/windows/svc/eventlog"
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+// stderr handle is invalid when run as service
+var logWriter = io.MultiWriter(ioutil.Discard)
```
petemoore: What is the reason to create an `io.MultiWriter` that wraps two writers, where one of them is always `ioutil.Discard`?

multiuser_service_windows.go
```diff
@@ -0,0 +1,295 @@
...
+		return err
+	}
+	log.Printf("Successfully configured service %q.", name)
+	err = configureEventlogSource(name, exePath)
+	if err != nil {
+		return err
+	}
+	log.Printf("Successfully configured eventlog source %q.", name)
+
+	// start service manually in order to fail fast
```
petemoore: When we create a machine image with generic-worker on it, we typically want to install it without starting it up (since it is just the machine image that real workers will use, and is not a real worker itself), so I think we don't want to start it up here, otherwise it would run on the machine which is not a worker, but it used to create the machine image for the actual workers.

helper_test.go
milescrabill: I believe it does, yes. Certain tests, some of the time. Were we not able to before?

I changed this to match the signature of the function, which I had changed before running the tests (yikes!) :P

simple_docker.go
milescrabill: Yeah, I added this to match `install()` but I don't think either are needed now that we're doing things in `platformTargets()`

usage_windows.go
milescrabill: What would the behavior be if that directory doesn't exist?

milescrabill: I think I'm going to change the behavior here.

If no `--working-directory` is specified, we default to the parent directory of the generic-worker executable. If one is specified, we use it.

multiuser_service_windows.go
petemoore:
```suggestion
var logWriter = io.MultiWriter()
```

multiuser_service_windows.go
```diff
@@ -0,0 +1,299 @@
...
+	"golang.org/x/sys/windows/svc/mgr"
+
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+// stderr handle is invalid when run as service
+var logWriter = io.MultiWriter(ioutil.Discard)
+
+func init() {
+	// TODO make this work with --working-directory
```
petemoore: I think that is done now, isn't it?

multiuser_service_windows.go
```diff
@@ -282,6 +284,8 @@ func removeService(name string) error {
 		return fmt.Errorf("service %s is not installed", name)
 	}
 	defer s.Close()
+	// throw away this error
```
petemoore: Can you add a note to say why we ignore the error?

multiuser_service_windows_test.go
```diff
@@ -78,10 +77,12 @@ func TestConfigureAndRemoveService(t *testing.T) {
 	}
 	elog.Close()
 
-	// this is unrealistic, usually the service is marked for deletion
+	// usually the service is marked for deletion
 	// but not actually removed until reboot
 
-	// TODO
+	cleanupService(t, name)
```
petemoore: What made you decide to move this out of the defer block?

multiuser_windows.go
```diff
@@ -447,25 +448,33 @@ func platformTargets(arguments map[string]interface{}) ExitCode {
...
+			log.Printf("failed to remove service: %v", err)
 			return CANT_REMOVE_GENERIC_WORKER
 		}
 	case arguments["run-service"]:
 		cwd := convertNilToEmptyString(arguments["--working-directory"])
-		if cwd != "" {
-			os.Chdir(cwd)
+		// default to generic-worker executable parent dir
+		if cwd == "" {
+			dir := path.Dir(os.Args[0])
```
petemoore:
```suggestion
			dir := filepath.Dir(os.Args[0])
```

usage_windows.go
```diff
@@ -62,7 +62,7 @@ func platformCommandLineParameters() string {
     --service-name SERVICE-NAME             The name that the Windows service should be
                                             installed under. [default: Generic Worker]
     --working-directory DIRECTORY           The working directory the Generic Worker
-                                            service will use. [default: C:\Windows\system32]`
+                                            service will use.`
```
petemoore: Can we determine the system default at runtime and include it in the usage text?

multiuser_service_windows_test.go
milescrabill: The commented out blocks below, if uncommented, would happen before a deferred call to `cleanupService()`

usage_windows.go
milescrabill: I think defaulting to the parent directory of the generic worker binary is a reasonable default. It's also trivial to implement in a cross platform manner.

multiuser_service_windows.go
milescrabill: If this fails it's essentially because something went wrong interacting with the service, i.e. it's not running, it didn't respond, etc. We don't care about an error in this scenario, we're just trying to delete it.

multiuser_service_windows.go
milescrabill: This always changes the working directiry to the parent directory of the generic worker binary.

multiuser_service_windows.go
petemoore: Perfect. Could you include that as a code comment?

helper_windows_test.go
```diff
@@ -116,3 +116,9 @@ func copyTestdataFileTo(src, dest string) []string {
 func singleCommandNoArgs(command string) []string {
 	return []string{command}
 }
+
+// this is opt-out so that we don't skip tests by default
+func shouldRunAdminTests() bool {
+	_, ok := os.LookupEnv("SKIP_ADMINISTRATOR_TESTS")
```
petemoore: Now we've landed #218 we should probably call this `GW_SKIP_<something>` and update the `README.md` to explain the purpose of this extra env variable. How about something like `GW_SKIP_SERVICE_INSTALLATION_TESTS`?

milescrabill: I renamed it to `GW_SKIP_ADMIN_REQUIRED_TESTS` to keep with the spirit of skipping administrator privileged tests.

usage_windows.go
petemoore: That's fine, but we should describe what the default is in the docs.

petemoore:
@milescrabill This has kind of bitrotted. Should we close it, or do you want to return to it?

PR #216, opened 2019-07-09
Bug 1564361 - log system errors when not able to check whether chain of trust key is readable by task user

petemoore:
See [bug 1564361](https://bugzil.la/1564361) for context.

### DIFF
--- a/chain_of_trust.go
+++ b/chain_of_trust.go
@@ -4,7 +4,10 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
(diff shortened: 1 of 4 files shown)


petemoore:
> This looks good to me, I'm wondering why the `Result` and `Command` definitions are in a file for the `simple` engine in `process` rather than a file that makes it clear they're used across engines.
> 
> Is that the pattern we've been using?

`Command` and `Result` are engine specific, for example to execute a command in a docker container is different to executing a command on the host operating system as a task user.

PR #217, opened 2019-07-10
Bug 1564819 - support docker engine on macOS

petemoore:
See [bug 1564819](https://bugzil.la/1564819) for context. Thanks!


.taskcluster.yml
```diff
@@ -694,3 +694,75 @@ tasks:
           dep check
       cache:
         generic-worker-checkout: /go/src
+
+
+  ##########################################################
+  ################ macOS amd64 Docker Build ################
+  ##########################################################
+
+  - provisionerId: pmoore-manual
```
milescrabill: Will this only work for your iMac?

petemoore: No, anything I manully provisioner I've been giving `"provisionerId": "pmoore-manual"` - so this covers a few different [worker types](https://tools.taskcluster.net/provisioners/pmoore-manual/worker-types), that themselves cover a few different workers.

The `provisionerId` essentially just partitions worker types into distinct projects.

milescrabill: Nice! Thanks for that link :)

PR #219, opened 2019-07-10
Bug 1564852 - enable linux multiuser builds, but don't test or release them

petemoore:
See [bug 1564852](https://bugzil.la/1564852) for context. Thanks!


petemoore:
> This looks good and fairly straightforward. I'm noticing that we have go versions hardcoded in `.taskcluster.yml` - what's the upgrade plan for those? Is there a way we could use something pseudo-latest?

That's a good question! At the moment, we intentionally stay on go 1.10 because we had an issue when upgrading to go 1.11 (needing gcc installed on Windows) and we haven't got to the root of that issue (understanding why we didn't need it before) nor added gcc to the workers so we aren't affected by it. It might not be a bad idea to take psuedo-latest, but at the moment, we pin.

petemoore:
Note, the work to get the tests passing, and enabling linux/amd64 multiuser releases is in #231.

PR #221, opened 2019-07-11
Bug 1565215 - include timezone in log timestamps and set to UTC

petemoore:
See [bug 1565215](https://bugzil.la/1565215) for context. Thanks!

### DIFF
--- a/logging.go
+++ b/logging.go
@@ -1,0 +1,21 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"time"
+)
+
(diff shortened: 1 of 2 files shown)


petemoore:
Hey Miles, since we put #214 on ice, maybe we can land this after all?

I think this is a small enough change that it doesn't require using a different logging library at this point. I think if at some point we want to overhaul logging, that would be a good time to do it, but this is a pretty small change that avoids the need for a dependency on another third party library.

PR #222, opened 2019-07-11
Bug 1565338 - avoid worker-state pollution between tests

petemoore:
See [bug 1565338](https://bugzil.la/1565338) for context. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,5 +46,7 @@ generic-worker.config
 directory-caches.json
 file-caches.json
 tasks-resolved-count.txt
+current-task-user.json
+next-task-user.json
 /revision.txt
 /generic-worker-*


petemoore:
I forgot this won't work, since we intentionally create `next-task-user.json` e.g. [here](https://github.com/taskcluster/generic-worker/blob/a1ff3e2e46903996823a9c19115af4207d672234/.taskcluster.yml#L72) in order that when the worker starts up in the test, it sees that a task user already exists, and that it matches the interactive task user.

Therefore, we are already taking care of preparing `next-task-user.json` and `current-task-user.json` correctly before running tests.

Therefore I'll limit this change to just the `.gitignore` changes.

PR #224, opened 2019-07-15
Bug 1566159 - determine uid/gid of current user by executing /usr/bin/id

petemoore:
See [bug 1566159](https://bugzil.la/1566159) for context.

### DIFF
--- a/fileutil/fileutil_posix.go
+++ b/fileutil/fileutil_posix.go
@@ -14,6 +14,8 @@ import (
 // filepaths, with 0600 file permissions.
 func SecureFiles(filepaths []string) (err error) {
 	// Use /usr/bin/id rather than user.Current due to https://bugzil.la/1566159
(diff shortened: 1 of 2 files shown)


fileutil/fileutil_posix.go
```diff
@@ -3,26 +3,35 @@
...
+	"strings"
 )
 
-// SecureFiles takes ownership of files, and then give them 0600 file permissions
+// SecureFiles makes the current user/group the owner of all files in
+// filepaths, with 0600 file permissions.
 func SecureFiles(filepaths []string) (err error) {
-	var currentUser *user.User
-	currentUser, err = user.Current()
+	// Use /usr/bin/id rather than user.Current due to https://bugzil.la/1566159
```
milescrabill: Might be worth mentioning what we talked about re: potentially switching to CGO in the future in a comment here

PR #225, opened 2019-07-16
Bug 1563220: switch to using docker client CLI

milescrabill:


process/docker.go
```diff
@@ -62,60 +48,41 @@ func (c *Command) String() string {
...
-	if r.SystemError != nil {
-		return
-	}
 
-	r.SystemError = c.cli.ContainerStart(c.ctx, c.resp.ID, types.ContainerStartOptions{})
-	if r.SystemError != nil {
+	image := "ubuntu"
+
+	// TODO scary injection potential here
+	cmd := exec.CommandContext(c.ctx, "/usr/bin/docker", append([]string{"run", image}, c.cmd...)...)
```
petemoore: We should probably make the docker executable a configuration setting (e.g. on my mac it is `/usr/local/bin/docker`).

petemoore: Don't we need to pull the docker image before using it?

petemoore: Let's also log any commands we run to the worker log.

milescrabill: We're logging commands now. Images get pulled as part of the `run` command, and are kept on disk for future use (not that we support that).

PR #226, opened 2019-07-17
Bug 1559777 - support deploying generic-worker linux workers to EC2

petemoore:
See [bug 1559777](https://bugzil.la/1559777) for context.


petemoore:
Hi @milescrabill, it will certainly be easiest to review this commit-by-commit, rather than looking at everything at once. The largest commit just migrates the format of the existing worker type definitions from the previous format to the new format, which would otherwise overshadow the other changes.

worker_types/gwci-linux/userdata
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
milescrabill: this script is a good start but could use some love for reliability purposes.

to start, sprinkle a little of this:

```
#!/bin/bash
set -e
set -x

function retry {
  set +e
  local n=0
  local max=3
  while true; do
    "$@" && break || {
      if [[ $n -lt $max ]]; then
        ((n++))
        echo "Command failed. Attempt $n/$max:"
      else
        echo "Failed after $n attempts."
        exit 1
      fi
    }
  done
  set -e
}

exec &> /var/log/userdata.log
start_time="$(date '+%s')"
# do stuff
end_time="$(date '+%s')"
echo "UserData execution took: $(($end_time-$start_time)) seconds"
```

You can use `retry` like so: `retry apt install -y docker-ce`

worker_types/gwci-linux/userdata
```diff
@@ -0,0 +1,31 @@
...
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable"
+apt update
+apt-cache policy docker-ce | grep -qF download.docker.com
+apt install -y docker-ce
+sleep 5
+systemctl status docker | grep "Started Docker Application Container Engine"
+usermod -aG docker ubuntu
+id -nG | grep docker
+curl -L https://github.com/taskcluster/generic-worker/releases/download/v15.1.2/generic-worker-simple-linux-amd64 > generic-worker
```
milescrabill: let's pull version numbers out into environment variables so they're easier to see

worker_types/gwci-linux/userdata
```diff
@@ -0,0 +1,31 @@
...
+curl -L https://github.com/taskcluster/taskcluster-proxy/releases/download/v5.1.0/taskcluster-proxy-linux-amd64 > taskcluster-proxy
+curl -L https://github.com/taskcluster/livelog/releases/download/v1.1.0/livelog-linux-amd64 > livelog
+chmod a+x generic-worker taskcluster-proxy livelog
+chown -R ubuntu:ubuntu /home/ubuntu/generic-worker
+./generic-worker --version
+./generic-worker new-ed25519-keypair --file ed25519.key
+echo '@reboot PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /home/ubuntu/generic-worker/generic-worker run --configure-for-aws --config /home/ubuntu/generic-worker/generic-worker.config >> /home/ubuntu/generic-worker/generic-worker.log 2>&1' | crontab -u ubuntu -
+# apt install -y gnome
```
milescrabill: are we keeping these commented out bits for now?

worker_types/gwci-linux/userdata
petemoore: I like it!

worker_types/gwci-linux/userdata
petemoore: I guess I could put them somewhere else... I commented them out because I suspect at some point we'll want to run stuff inside an actual desktop environment, and I didn't want to lose the steps I'd already worked out... But maybe I'm just making a mess by leaving them in there.

petemoore:
We shouldn't land this yet - the shutdown command when run from generic-worker is giving an exit code of 1 and not working, and therefore the machines are not shutting down when they are idle...

I've created [bug 1567073](https://bugzil.la/1567073) to find out why the shutdown command is failing, as we currently don't log the stdout/stderr from the shutdown command. 😢

PR #227, opened 2019-07-18
Bug 1567073 - log all commands that are executed on the host

petemoore:
See [bug 1567073](https://bugzil.la/1567073) for context.

I created a separate package (`host`) to avoid some cyclic dependencies we would have otherwise had. The idea is, this should be anything you want to do to interact with the host operating system, rather than interacting with e.g. a container/sandbox. At the moment it just contains an existing `RunCommands` function that I moved straight over from `runtime` package, and a new `RunCommand` utility method to log the command about to be run, run it, check if it fails, and if so log the combined stderr/stdout, and returns the command output and error. For convenience, it converts the output from `[]byte` to `string` since in 99% of cases that is what we want.

The other changes are just to use it, so we have a single function where we execute commands, that takes care of logging them and reporting any errors.

Note, since the function _logs the command output if the command fails_, I am no longer logging the same output from the caller. Instead the caller just typically logs the error (e.g. `exit code 1`). I chose to log the command output _inside_ the method, so that it was just in one place, rather than anything that gets an error doing exactly the same thing. However, since `error` types are typically passed back up the stack and logged at the top level where the error is handled, I intentionally am not logging the error inside `host.RunCommand` so that we don't get it logged twice unnecessarily. Often more context is added to the error message as it gets passed back up the stack, so better for _callers_ to log their own errors, rather than the `RunCommand` method that doesn't know if the `error` is significant or not.

### DIFF
--- a/host/host.go
+++ b/host/host.go
@@ -1,0 +1,69 @@
+// Package host provides facilities for interfacing with the host operating
+// system, logging activities performed.
+package host
+
+import (
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/taskcluster/shell"
+)
+
+// Run is equivalent to exec.Command(name, arg...).Run() but with logging.
+func Run(name string, arg ...string) (err error) {
+	_, err = RunCommand(exec.Command(name, arg...))
+	return
+}
+
+// CombinedOutput is equivalent to exec.Command(name, arg...).CombinedOutput()
+// but with logging.
+func CombinedOutput(name string, arg ...string) (combinedOutput string, err error) {
+	return RunCommand(exec.Command(name, arg...))
+}
+
+// RunBatch calls Run for each command in commands, in sequence. If allowFail
+// is false it will return if an error is returned from Run. All errors are
+// logged regardless of allowFail. The returned error is the result of the last
+// Run call.
+func RunBatch(allowFail bool, commands ...[]string) (err error) {
+	for _, command := range commands {
+		err = Run(command[0], command[1:]...)
+		if err != nil {
+			log.Printf("%v", err)
+			if !allowFail {
+				return err
+			}
+		}
+	}
+	return err
+}
+
+// RunIgnoreError calls CombinedOutput(comand, args...). If errString is found
+// in the command output, found is true and err is nil.  Otherwise found is
+// false, and err is the error returned from CombinedOutput (possibly nil).
+func RunIgnoreError(errString string, command string, args ...string) (found bool, err error) {
+	output, err := CombinedOutput(command, args...)
+	if err != nil {
+		if strings.Contains(output, errString) {
+			return true, nil
+		}
+	}
+	return false, err
+}
+
+// RunCommand logs cmd.Args, calls cmd.CombinedOutput(), and if an error
+// occurs, logs the command output. It does not log the error, it is expected
+// that the caller takes care of logging error, if required. The caller is not
+// expected to log the command output in the case of failure, since this
+// function has already done that. The combined output is cast to a string and
+// returned together with the error.
+func RunCommand(cmd *exec.Cmd) (combinedOutput string, err error) {
+	log.Print("Running command: " + shell.Escape(cmd.Args...))
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		log.Print("Error running command:")
+		log.Print(string(out))
+	}
+	return string(out), err
+}
--- a/multiuser_posix.go
+++ b/multiuser_posix.go
@@ -6,9 +6,9 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"os/exec"
 	"path/filepath"
 
+	"github.com/taskcluster/generic-worker/host"
 	"github.com/taskcluster/generic-worker/process"
 	"github.com/taskcluster/generic-worker/runtime"
 	gwruntime "github.com/taskcluster/generic-worker/runtime"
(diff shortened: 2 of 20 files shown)


host/host.go
```diff
@@ -0,0 +1,38 @@
...
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/taskcluster/shell"
+)
+
+func RunCommand(cmd *exec.Cmd) (combinedOutput string, err error) {
+	log.Print("Running command: " + shell.Escape(cmd.Args...))
+	out, err := cmd.CombinedOutput()
```
milescrabill: We're not actually running `shell.Escape()` on the `cmd.Args` that gets run here, just in the log printout - it seems like we should escape the actual commandline, right?

host/host.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		log.Print("Error running command:")
+		log.Print(string(out))
+	}
+	return string(out), err
+}
+
+func RunCommands(allowFail bool, commands ...[]string) error {
```
milescrabill: I'm a little confused at the intention here - this plural command seems like it would be a loop around multiple invocations of `RunCommand`, but it takes a slice of strings as its arguments, not `*exec.Cmd`s. Per my comment above, I think we could keep this accepting a slice of strings and convert it to wrap a loop around invocations of `RunCommand`.

host/host.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"log"
+	"os"
+	"os/exec"
+
+	"github.com/taskcluster/shell"
+)
+
+func RunCommand(cmd *exec.Cmd) (combinedOutput string, err error) {
```
milescrabill: I think this should take a slice of strings (preventing the need to call this function like `RunCommand(exec.Command([]{"path", "arg1", "arg...", "argn"}))`) for convenience

petemoore: Yeah, maybe. I was torn between the two. The reason I chose to pass in the command, is because it gives the caller the freedom to set env variables or other settings inside the command. Sometimes we set auth tokens, or env vars, in the commands we run, so I figured this was the way to provide the most freedom, but it does come at the cost of making the call site a little bit more clunky.

host/host.go
petemoore: This method was just moved wholesale from runtime package, I totally agree we should clean up all the utility methods we have around task execution, I just didn't do that in this PR. :-) This is just taking it as it was, and relocating it. But yes, it is pretty ugly, so I can follow up with another PR to address that, it is just a bit more scope creep from the original bug.

host/host.go
petemoore: The `shell.Escape` function just outputs command arguments in a shell escaped fashion, such that you could copy/paste the output into a bash shell, and the arguments would be escaped correctly. `cmd.Args` is the component of the command that contains the arguments (including the program itself), as opposed to e.g. the env vars or other state associated with the command.

host/host.go
milescrabill: I take this back - it's worth allowing the caller the flexibility to set up an `*exec.Cmd` with things set on it.

host/host.go
milescrabill: I think ideally we prevent the kinds of shell escapes that can allow user submitted payloads to do scary things on workers. Something like `shell.Escape()` can be used to do that.

These kinds of scary things: https://github.com/taskcluster/generic-worker/blob/master/process/docker.go#L62-L63

host/host.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"log"
+	"os/exec"
+	"strings"
+
+	"github.com/taskcluster/shell"
+)
+
+// Run is equivalent to exec.Command(name, arg...).Run() but with logging.
+func Run(name string, arg ...string) (err error) {
```
milescrabill: Uses of this will throw away stdout/stderr and not log successful command output from either, is that what we want?

host/host.go
```diff
@@ -0,0 +1,69 @@
...
+// but with logging.
+func CombinedOutput(name string, arg ...string) (combinedOutput string, err error) {
+	return RunCommand(exec.Command(name, arg...))
+}
+
+// RunBatch calls Run for each command in commands, in sequence. If allowFail
+// is false it will return if an error is returned from Run. All errors are
+// logged regardless of allowFail. The returned error is the result of the last
+// Run call.
+func RunBatch(allowFail bool, commands ...[]string) (err error) {
```
milescrabill: The normal flow for `RunBatch` will double print output in the case of an error.

`RunBatch()` (prints errors) => `Run()` => `RunCommand()` (prints errors)

multiuser_posix.go
```diff
@@ -174,19 +153,19 @@ func MkdirAllTaskUser(dir string, perms os.FileMode) (err error) {
...
 }
 
-func makeFileOrDirReadWritableForUser(recurse bool, dir string, user *gwruntime.OSUser) ([]byte, error) {
+func makeFileOrDirReadWritableForUser(recurse bool, dir string, user *gwruntime.OSUser) error {
 	if recurse {
-		return exec.Command("/usr/sbin/chown", "-R", user.Name+":staff", dir).CombinedOutput()
+		return host.Run("/usr/sbin/chown", "-R", user.Name+":staff", dir)
 	}
-	return exec.Command("/usr/sbin/chown", user.Name+":staff", dir).CombinedOutput()
+	return host.Run("/usr/sbin/chown", user.Name+":staff", dir)
```
milescrabill: The message here is that we've already logged the error, so we don't care about having stdout/stderr here - I like that, it seems like the right approach.

multiuser_windows.go
```diff
@@ -66,12 +47,7 @@ func deleteDir(path string) error {
 	log.Print("WARNING: could not delete directory '" + path + "' with os.RemoveAll(path) method")
 	log.Printf("%v", err)
 	log.Print("Trying to remove directory '" + path + "' via del command...")
-	err = runtime.RunCommands(
-		false,
-		[]string{
-			"cmd", "/c", "del", "/s", "/q", "/f", path,
-		},
-	)
+	err = host.Run("cmd", "/c", "del", "/s", "/q", "/f", path)
```
milescrabill: This is cleaner!

process/multiuser_posix.go
```diff
@@ -27,21 +29,36 @@ func NewPlatformData(currentUser bool) (pd *PlatformData, err error) {
 func TaskUserPlatformData() (pd *PlatformData, err error) {
 	user, err := runtime.InteractiveUsername()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Could not determine interactive username: %v", err)
+	}
+
+	id := func(idName string, idOpts string) (uint32, error) {
```
milescrabill: The way this is called is confusing to me because it looks like `"UID"` or `"GID"` are being passed as commands to the `id` command itself. I'd almost rather have separate functions split out, as clunky as that sounds. The `host` package could be a good place for `UserID()` and `UserGID()` or similar.

host/host.go
petemoore: Yes, the use case here is when you only care about the output if something goes wrong. If you are interested in the output from a successful command, there is the `CombinedOutput` method which is the same, but retains the output.

process/multiuser_posix.go
petemoore: I see what you mean. I guess I could pass the entire command through, rather than just the opts. It felt clunky to me too when I wrote the method, but I also disliked copy/pasting the code when it was essentially the same. 

Maybe if I pass the entire command through it will look cleaner...

multiuser_windows.go
petemoore: Thanks!

host/host.go
petemoore: No, `RunCommand()` logs the output of the failed command, whereas RunBatch() displays the (go) error (e.g. 'exit code 65'). The `RunCommand` docs explain that the caller should log the error, and it will log the command output.

multiuser_posix.go
petemoore: Thanks!

PR #228, opened 2019-07-18
WIP: Bug 1502337 - Support specifying image to be fetched in task payload

milescrabill:
See [bug 1502337](https://bugzil.la/1502337) for context.

### DIFF
--- a/docker_posix.yml
+++ b/docker_posix.yml
@@ -221,8 +221,8 @@ properties:
           minimum: 1
   image:
     title: Docker image.
-    description: Image to use for the task.  Images can be specified as an image tag as
-      used by a docker registry, or as an object declaring type and name/namespace
+    description: Image to use for the task.  Images can be specified as an image tag
+      as used by a docker registry, or as an object declaring type and name/namespace
     oneOf:
     - title: Docker image name
       type: string
@@ -238,6 +238,36 @@ properties:
       required:
       - type
       - name
+    - type: object
+      title: Indexed docker image
+      properties:
+        type:
+          type: string
+          enum:
+          - indexed-image
+        namespace:
+          type: string
+        path:
+          type: string
+      required:
+      - type
+      - namespace
+      - path
+    - type: object
+      title: Docker image artifact
+      properties:
+        type:
+          type: string
+          enum:
+          - task-image
+        taskId:
+          type: string
+        path:
+          type: string
+      required:
+      - type
+      - taskId
+      - path
 definitions:
   mount:
     title: Mount
--- a/docker_posix.yml
+++ b/docker_posix.yml
@@ -219,6 +219,25 @@ properties:
           title: Exit codes
           type: integer
           minimum: 1
+  image:
+    title: Docker image.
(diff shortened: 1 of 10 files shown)


docker_posix.yml
```diff
@@ -219,6 +219,55 @@ properties:
...
+      properties:
+        type:
+          type: string
+          enum:
+          - task-image
+        taskId:
+          type: string
+        path:
+          type: string
+      required:
```
petemoore: also `additionalProperties: false` here

docker_posix.yml
```diff
@@ -219,6 +219,55 @@ properties:
...
+      properties:
+        type:
+          type: string
+          enum:
+          - indexed-image
+        namespace:
+          type: string
+        path:
+          type: string
+      required:
```
petemoore: also `additionalProperties: false` here

docker_posix.yml
```diff
@@ -219,6 +219,55 @@ properties:
...
+    - type: object
+      title: Named docker image
+      properties:
+        type:
+          type: string
+          enum:
+          - docker-image
+        name:
+          type: string
+      required:
```
petemoore: At the same level as `required:` you will also need `additionalProperties: false`, to disallow that other properties are included.

generated_docker_darwin.go
```diff
@@ -241,6 +272,40 @@ type (
...
+	//
+	//  	Path string `json:"path"`
+	//
+	//  	// Possible values:
+	//  	//   * "indexed-image"
+	//  	Type string `json:"type"`
+	//  }
+	//
+	// Additional properties allowed
+	IndexedDockerImage json.RawMessage
```
petemoore: Same here.

generated_docker_darwin.go
```diff
@@ -241,6 +272,40 @@ type (
...
+	//
+	//  	Name string `json:"name"`
+	//
+	//  	// Possible values:
+	//  	//   * "docker-image"
+	//  	Type string `json:"type"`
+	//  }
+	//
+	// Additional properties allowed
+	NamedDockerImage json.RawMessage
```
petemoore: Same here.

generated_docker_darwin.go
```diff
@@ -195,6 +217,15 @@ type (
...
 		Features FeatureFlags `json:"features,omitempty"`
 
+		// Image to use for the task.  Images can be specified as an image tag as used by a docker registry, or as an object declaring type and name/namespace
+		//
+		// One of:
+		//   * DockerImageName
+		//   * NamedDockerImage
+		//   * IndexedDockerImage
+		//   * DockerImageArtifact
+		Image json.RawMessage `json:"image,omitempty"`
```
petemoore: The reason this is generated as `json.RawMessage` is because `additionalProperties: false` is not specified.

petemoore (changes requested):
Looks like a good start, many thanks Miles!

petemoore:
@milescrabill Are you still working on this, or should we close it?

PR #229, opened 2019-07-19
Trivial logging changes

petemoore:
Not worthy of a bug, or probably even a review, but here it is anyway...

### DIFF
--- a/main.go
+++ b/main.go
@@ -838,7 +838,7 @@ func (task *TaskRun) kill() {
 			task.Info(string(output))
 		}
 		if err != nil {
-			log.Printf("WARNING - %v", err)
+			log.Printf("WARNING: %v", err)
 			task.Warnf("%v", err)
 		}
 	}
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -837,7 +837,7 @@ func (task *TaskRun) kill() {
 			task.Info(string(output))
 		}
 		if err != nil {
-			log.Printf("WARNING - %v", err)
+			log.Printf("WARNING: %v", err)
```
petemoore: This change is just to make our warning log messages consistent.

runtime/runtime_darwin.go
```diff
@@ -74,6 +74,7 @@ func WaitForLoginCompletion(timeout time.Duration) error {
 	for time.Now().Before(deadline) {
 		username, err := InteractiveUsername()
 		if err != nil {
+			log.Printf("WARNING: Error checking for interactive user: %v", err)
```
petemoore: Weirdly in CI I saw that sometimes `InteractiveUsername()` was called twice, and I've no idea why it failed the first time, so this should help diagnose.

PR #230, opened 2019-07-24
Bug 1568471 - close HTTP response bodies when finished with them

petemoore:
See [bug 1568471](https://bugzil.la/1568471) for context. Thanks!

### DIFF
--- a/expose/proxy_test.go
+++ b/expose/proxy_test.go
@@ -189,10 +189,11 @@ func TestProxyHTTPWebsocket(t *testing.T) {
 
 	url := fmt.Sprintf("ws://127.0.0.1:%d", listenPort)
 	dialer := websocket.Dialer{}
-	ws, _, err := dialer.Dial(url, nil)
+	ws, resp, err := dialer.Dial(url, nil)
 	if err != nil {
 		t.Fatalf("client Dial: %s", err)
 	}
+	defer resp.Body.Close()
 	defer ws.Close()
 
 	err = ws.WriteMessage(websocket.BinaryMessage, []byte("ECHO"))
@@ -256,7 +257,11 @@ func TestProxyTCPPortNotRootPath(t *testing.T) {
 
 	url := fmt.Sprintf("ws://127.0.0.1:%d/random/path", listenPort)
 	dialer := websocket.Dialer{}
-	_, _, err = dialer.Dial(url, nil)
+	var resp *http.Response
+	_, resp, err = dialer.Dial(url, nil)
+	if err == nil {
+		resp.Body.Close()
+	}
 	assert.Equal(t, fmt.Errorf("websocket: bad handshake"), err, "should get bad handshake")
 }
 
@@ -291,10 +296,11 @@ func TestProxyTCPPort(t *testing.T) {
 	// connect to that ws proxy
 	url := fmt.Sprintf("ws://127.0.0.1:%d/", listenPort)
 	dialer := websocket.Dialer{}
-	ws, _, err := dialer.Dial(url, nil)
+	ws, resp, err := dialer.Dial(url, nil)
 	if err != nil {
 		t.Fatalf("Dial: %s", err)
 	}
+	defer resp.Body.Close()
 
 	err = ws.WriteMessage(websocket.BinaryMessage, []byte("Hello"))
 	if err != nil {
@@ -351,10 +357,11 @@ func TestProxyTCPPortTCPClose(t *testing.T) {
 	// connect to that ws proxy
 	url := fmt.Sprintf("ws://127.0.0.1:%d/", listenPort)
 	dialer := websocket.Dialer{}
-	ws, _, err := dialer.Dial(url, nil)
+	ws, resp, err := dialer.Dial(url, nil)
 	if err != nil {
 		t.Fatalf("Dial: %s", err)
 	}
+	defer resp.Body.Close()
 
 	messageType, payload, err := ws.ReadMessage()
 	if err != nil {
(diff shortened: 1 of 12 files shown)


mounts.go
```diff
@@ -800,11 +800,13 @@ func downloadURLToFile(url, contentSource, file string, task *TaskRun) (sha256 s
...
 		return resp, nil, nil
 	}
-	_, _, err = httpbackoff.Retry(retryFunc)
+	var resp *http.Response
+	resp, _, err = httpbackoff.Retry(retryFunc)
 	if err != nil {
 		task.Errorf("[mounts] Could not fetch from %v into file %v: %v", contentSource, file, err)
 		return
 	}
+	defer resp.Body.Close()
```
milescrabill: If we're not using `resp` we should just close it here.

supersede_test.go
```diff
@@ -66,7 +66,8 @@ func TestSupersede(t *testing.T) {
 			ensureResolution(t, taskID, "completed", "completed")
 		} else {
 			ensureResolution(t, taskID, "exception", "superseded")
-			x, _, _, _ := getArtifactContent(t, taskID, "public/superseded-by.json")
+			x, resp, _, _ := getArtifactContent(t, taskID, "public/superseded-by.json")
+			defer resp.Body.Close()
```
milescrabill: If we're not using `resp` we should just close it here.

worker_types/all-worker-types/main.go
```diff
@@ -105,10 +105,9 @@ func (wtf *WorkerTypeFetcher) fetch(workerType string) (err error) {
 		return
 	}
 	defer func() {
+		err2 := resp.Body.Close()
```
milescrabill: What's the logic here? Not shadowing `err`?

worker_types/transform-occ/main.go
```diff
@@ -219,10 +219,11 @@ func FilenameFromURL(url string, extension string) (filename string) {
 			return nil
 		},
 	}
-	_, err := client.Head(url)
+	resp, err := client.Head(url)
 	if err != nil {
 		log.Fatal("Could not reach URL " + url)
 	}
+	defer resp.Body.Close()
```
milescrabill: If we're not using `resp` we should just close it here.

worker_types/all-worker-types/main.go
petemoore: `err` is returned by the function. If an error has already occurred before the defer function is called (`err != nil`) then we should return that existing error (`err`). However, if `err == nil` (i.e. no error has occurred prior to calling the defer function) but `err2 != nil` (i.e. closing the response body causes an error) then we should return `err2` instead. Whatever we set `err` to will be what gets returned, so setting `err` to `err2` in this case means that we will return the error from closing the response body if one happens, but previously no error had occurred. If an error had already occurred, and we fail to close the response body, we'll return the original error, not the error about closing the response body.

expose/proxy_test.go
```diff
@@ -189,10 +189,11 @@ func TestProxyHTTPWebsocket(t *testing.T) {
 
 	url := fmt.Sprintf("ws://127.0.0.1:%d", listenPort)
 	dialer := websocket.Dialer{}
-	ws, _, err := dialer.Dial(url, nil)
+	ws, resp, err := dialer.Dial(url, nil)
 	if err != nil {
 		t.Fatalf("client Dial: %s", err)
 	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
petemoore: Maybe. The reason I preferred to defer is that imagine someone comes alone and needs to do something with `resp`. It would be quite easy for them to add code directly before the `resp.Body.Close()` line that did something with `resp`, and not notice any problems. But they will have introduced a bug, because they could easily not notice that they need to change `resp.Body.Close()` to `defer resp.Body.Close()`. By having the `defer` statement there from the beginning, there is no risk that this happens. What is the disadvantage of closing `resp` at the end of the function? I don't think it comes at much of an additional cost, but helps to future-proof the code a little.

I can change if you really want, but I think both versions have their merits.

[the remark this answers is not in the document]
milescrabill: I think your point is reasonable. Reading from a closed reader is a panic, so in this hypothetical you'd notice, but I've come around.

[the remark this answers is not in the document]
petemoore: Just to be clear, I mean a slightly different scenario. Let's say the code looks initially like this, without the `defer` statement:

```
	ws, resp, err := dialer.Dial(url, nil)
	if err != nil {
		t.Fatalf("client Dial: %s", err)
	}
	resp.Body.Close()
```

A new requirement comes in, that makes Tuesdays special, such that no further work needs to be done on a Tuesday. Perhaps something like the following gets added:

```
	ws, resp, err := dialer.Dial(url, nil)
	if err != nil {
		t.Fatalf("client Dial: %s", err)
	}
	// Bug 1234567: tuesdays are special
	if resp.Header.get("Day-of-Week") == "Tuesday" {
		return
	}
	resp.Body.Close()
```

This change won't result in a runtime panic since the bug is slightly more subtle: on Tuesdays, the response body won't be closed. Of course, they may notice that they need to change the line below to be in a defer block, but it is easy to miss.

If it started life as this:

```
	ws, resp, err := dialer.Dial(url, nil)
	if err != nil {
		t.Fatalf("client Dial: %s", err)
	}
	defer resp.Body.Close()
```

then the change is more likely to be applied as:

```
	ws, resp, err := dialer.Dial(url, nil)
	if err != nil {
		t.Fatalf("client Dial: %s", err)
	}
	defer resp.Body.Close()
	if resp.Header.get("Day-of-Week") == "Tuesday" {
		return
	}
```

Of course it is still possible that the code gets added before the `defer` statement (which would also be wrong), so having the `defer` from the start isn't foolproof, but I think it is easier to spot when adding the new hypothetical feature.

It can also be that I am overthinking the problem! :-)

PR #231, opened 2019-07-24
Bug 1499054 - enable multiuser CI tests on Linux

petemoore:
Originally tackled in #211, we closed that PR in favour of splitting the work into two parts.

The first part was #219 which took care of writing skeleton stubs for linux, and enabling multiuser builds (not not tests) for linux in the generic-worker CI.

This second PR is for converting the skeleton stubs into working code, and enabling the linux multiuser tests in the generic-worker CI.

See [bug 1499054](https://bugzil.la/1499054) for more details.


worker_types/gwci-linux/userdata
```diff
@@ -46,22 +47,32 @@ sleep 5
...
-
 # configure generic-worker to run on boot
-echo '@reboot PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /home/ubuntu/generic-worker/generic-worker run --configure-for-aws --config /home/ubuntu/generic-worker/generic-worker.config >> /home/ubuntu/generic-worker/generic-worker.log 2>&1' | crontab -u ubuntu -
+echo '@reboot cd /var/local/generic-worker && PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin /usr/local/bin/generic-worker run --configure-for-aws --config /etc/generic-worker/config >> /var/log/generic-worker.log 2>&1' | crontab -
+
+retry apt install -y ubuntu-desktop ubuntu-gnome-desktop
+
+# See
+#   * https://console.aws.amazon.com/support/cases#/6410417131/en
+#   * https://bugzilla.mozilla.org/show_bug.cgi?id=1499054#c12
```
milescrabill: Interesting, never seen this before.

worker_types/gwci-linux-beta/userdata
```diff
@@ -0,0 +1,93 @@
...
+retry curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable"
+retry apt update
+apt-cache policy docker-ce | grep -qF download.docker.com
+retry apt install -y docker-ce
+sleep 5
+systemctl status docker | grep "Started Docker Application Container Engine"
+usermod -aG docker ubuntu
+
+# build generic-worker from bug1499054 branch
```
milescrabill: Will we continue to build off this branch post-merge?

runtime/runtime_linux.go
```diff
@@ -2,39 +2,220 @@ package runtime
...
+	}
+	return lines[0], nil
+}
+
+type sectionHandler interface {
+	lineInsideSection(line string)
+	lineOutsideSection(line string)
+}
+
+func iniFileLineHandler(iniFile string, section string, sh sectionHandler) (err error) {
```
milescrabill: Does this kind of logic make sense to handle in a subpackage?

multiuser_posix.go
```diff
@@ -153,11 +156,23 @@ func MkdirAllTaskUser(dir string, perms os.FileMode) (err error) {
 	return nil
 }
 
-func makeFileOrDirReadWritableForUser(recurse bool, dir string, user *gwruntime.OSUser) error {
+func makeFileOrDirReadWritableForUser(recurse bool, fileOrDir string, user *gwruntime.OSUser) error {
 	if recurse {
-		return host.Run("/usr/sbin/chown", "-R", user.Name+":staff", dir)
+		switch runtime.GOOS {
+		case "darwin":
+			return host.Run("/usr/sbin/chown", "-R", user.Name+":staff", fileOrDir)
```
milescrabill: Is there a reason why we're not using the `os` package here? If so, let's document it.

https://golang.org/pkg/os/#Chown

runtime/runtime_darwin.go
```diff
@@ -26,8 +26,6 @@ func (user *OSUser) CreateNew(okIfExists bool) (err error) {
 		fullname="${0}"
 		password="${1}"
 		echo "Creating user '${username}' with home directory '${homedir}' and password '${password}'..."
-		maxid=$(dscl . -list '/Users' 'UniqueID' | awk '{print $2}' | sort -ug | tail -1)
```
milescrabill: It looks like this prevents us from creating a user with an id that is already taken. Is this no longer necessary?

runtime/runtime_linux.go
```diff
@@ -2,39 +2,220 @@ package runtime
...
+		}
+	}
+	return
 }
 
 func UserHomeDirectoriesParent() string {
 	return "/home"
 }
 
 func WaitForLoginCompletion(timeout time.Duration) error {
```
milescrabill: This timeout/sleep deadline works, though it might be more idiomatic to use a context here.

runtime/runtime_linux.go
```diff
@@ -2,39 +2,220 @@ package runtime
...
+		}
+		return nil
+	}
+	log.Print("Timed out waiting for user login")
+	return errors.New("No user logged in with console session")
 }
 
 func InteractiveUsername() (string, error) {
-	return "", errors.New("runtime.InteractiveUsername() not implemented on Linux")
+	gnomeSessionUserList, err := host.CombinedOutput("/bin/bash", "-c", "PROCPS_USERLEN=20 /usr/bin/w -s | /bin/grep gnome-[s]ession | /usr/bin/cut -f1 -d' '")
```
milescrabill: What exactly is this doing? We don't know the uid of the interactive user and have to grep for it via w? The interactive session has a separate uid than the logged in user?

milescrabill:
I don't want to block you on this, so I went with the approve, but I have some open questions.

runtime/runtime_darwin.go
petemoore: We used to calculate what the `uid` would be, since we knew it was the existing maximum `uid` + 1, and we would output it in the script. Then the calling code would parse the output and store the `uid`. We no longer need the `uid` so at some point we removed the code that output it and then parsed it, but I had not noticed that the code to predict the `uid` was still in there, so this was just a cleanup to remove that code. `uid` and `maxid` are never referenced anywhere else, so it is safe to remove these lines.

runtime/runtime_linux.go
petemoore: This is the way we determine the `uid` of the interactive user (at the moment we only support gnome graphical logins) so this is how we first find out which user is logged in. Note, generic-worker keeps a record of which user it expects to have a graphical login session (i.e. the task user it created that it reboots into) but this is how it checks that the user is said should auto-login *is* the user that is interactively logged in. If it discovers a different user has the graphical login session than the one and only user it expected to have a graphical login session, it will exit with an error message explaining this.

worker_types/gwci-linux-beta/userdata
petemoore: This is just the staging environment, so I think it is ok to leave it in this state, as it reflects what is currently deployed in staging. When we deploy something else to staging, I think we'll update it at that time, which will keep the code in sync with what is deployed there.

multiuser_posix.go
petemoore: ++ You are quite right - there was a reason we weren't able to use the `os` package implementation, and I should have documented why, because I have myself forgotten now! I will have a think about this and try to remember, or if I can't, I'll switch it to use `os` package, see why it breaks, switch it back and add an explanation why.

runtime/runtime_linux.go
petemoore: Fair point, I'll see if I can make that more idiomatic. ++

runtime/runtime_linux.go
petemoore: That's a good idea. I might slightly adapt/simplify it too at the same time.

worker_types/gwci-linux/userdata
petemoore: Yeah, I was a little baffled but the AWS support guy was really helpful.

petemoore:
> This looks good, but all of this shelling out and modifying system files feels strange to bake into the main codebase. Since we're already embedding bash in our go code, would it make more sense to write single purpose scripts and exec them from generic-worker?

So I have mixed feelings about this. The reason I like to have this stuff baked in is that you know it hasn't been tampered with. Both from a security perspective things are a little simpler (we don't need to make sure those other files are locked down and can't be interfered with by tasks) but also it reduces the risk that somebody operating the platform makes unexpected changes. I also really like that we ship a single binary, which is self-contained with no dependencies. As soon as it calls out to other files, we need to be sure it isn't looking in the wrong directory for them, or that they don't get copied to the wrong place etc etc. I definitely like the "trustability" with the current setup, that they do what we expect them to do.

The flip side is the code is a little less transparent. It is harder for people to customise the scripts, although, I think in general that is a good thing, as they have a very specific function. I think in the case that they contained a bug, we would want to ship a new release anyway, so I don't think there is a strong argument that they should be separate so that they can be updated outside the normal release/deployment workflow of the generic-worker program.

These are the reasons I chose to embed the scripts, which for the most part are usually only a couple of lines. We maybe could be more explicit about system dependencies (so it is known on each platform what the specific binary dependencies are). Maybe let's talk this through in person. I also see that intuitively it can feel against good design principles to embed bash scripts in a go binary, but I think logically it makes sense, even if intuitively it doesn't, but I'm open to discuss!

petemoore:
> I don't want to block you on this, so I went with the approve, but I have some open questions.

In this case I'm not in a huge rush to land, so I am happy to make sure we're both happy with the changes before I land, but I very much appreciate the sentiment, so many thanks. :-)

runtime/runtime_linux.go
petemoore: On reflection, `WaitForLoginCompletion` spans all platforms, so changing the method signature here would mean changing all platform implementations, which may be a bit of unnecessary scope creep for this PR, so I think I'll leave it as if for now, and another PR can overhaul the `WaitForLoginCompletion` method across all platforms (windows/linux/darwin) to be more idiomatic, if required.

gdm3/gdm3.go
```diff
@@ -0,0 +1,105 @@
+// package gdm3 provides functions for interfacing with Gnome Desktop Manager 3 on linux
```
milescrabill: This feels way cleaner to me, nice work!

gdm3/gdm3_test.go
```diff
@@ -0,0 +1,45 @@
+package gdm3
+
+import "testing"
+
+func TestSetAutoLogin(t *testing.T) {
```
milescrabill: 🙌

ini/ini.go
```diff
@@ -0,0 +1 @@
+package ini
```
milescrabill: Looks like this is an empty package, let's delete it

petemoore: Whoops, nice spot!

PR #232, opened 2019-07-25
Bug 1568782 - error artifacts do not have a content type

petemoore:
See [bug 1568782](https://bugzil.la/1568782) for context. Thanks!


petemoore:
Thanks!

PR #238, opened 2019-09-25
Fetch taskcluster test credentials from taskcluster-secrets

petemoore:
For generic-worker CI tests, we need a taskcluster client with a limited set of scopes. On some worker types, we previously burned in a client on the workers, but this has some disadvantages:

1) It means rotating the client requires updating all workers that use it, which can be some work
2) It is less clear / discoverable where the client comes from, from the task definition
3) It makes it harder for users that should be able to see the secret, to see it

By updating the CI tasks to consistently fetch the secret from taskcluster-secrets, it removes the need to burn it into workers, which can then also be used for other tasks too that shouldn't have access to this secret, if needed.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -30,9 +30,18 @@ tasks:
           - push
     scopes:
       - generic-worker:cache:generic-worker-checkout
+      - secrets:get:repo:github.com/taskcluster/generic-worker
     payload:
+      features:
+        taskclusterProxy: true
       maxRunTime: 3600
       command:
+        - >-
+          C:\cygwin\bin\wget.exe -q -O-
+          %TASKCLUSTER_PROXY_URL%/secrets/v1/secret/repo:github.com/taskcluster/generic-worker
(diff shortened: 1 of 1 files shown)


petemoore:
Note, there was an intermittent failure in the CI, which was solved by a task rerun. Therefore if you follow the link to the failed task, you'll see it resolved successfully in the rerun.

.taskcluster.yml
```diff
@@ -68,8 +77,9 @@ tasks:
         - go install -tags multiuser -v -ldflags "-X main.revision=%revision%" ./...
         - set CGO_ENABLED=1
         - set GORACE=history_size=7
-        - C:\generic-worker\generic-worker-test-creds.cmd
```
milescrabill: What did this do / why is it no longer needed?

petemoore: This was the file that set the credentials, that was burned into the worker image. With this PR, this file is now fetched from taskcluster-secrets.

PR #245, opened 2019-10-15
Bug 1588789 - log workerReady the first time we're ready to call queue.claimWork

petemoore:
See [bug 1588789](https://bugzil.la/1588789) for details.


petemoore:
Failures seem to be related to docker registry unavailability, rather than bugs in code, so I'll merge. Only docker engine had issues, and this isn't used in production, and the problem looks like it wasn't on our side anyway.

PR #247, opened 2019-10-18
Bug 1588834 - add support for AWS provider

petemoore:
Let's see how the CI runs with this. It is mostly shuffling code around. If the CI passes, we should have reasonable coverage for AWS Provisioner and GCP. We can add tests for AWS Provider, but I haven't done that yet.

See [bug 1588834](https://bugzil.la/1588834) for context.

### DIFF
--- a/awsprovisioner.go
+++ b/awsprovisioner.go
@@ -1,0 +1,85 @@
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/taskcluster/generic-worker/gwconfig"
+	"github.com/taskcluster/taskcluster-client-go/tcawsprovisioner"
+)
+
+var (
+	// for querying deploymentId
+	provisioner *tcawsprovisioner.AwsProvisioner
+)
+
+// taken from https://github.com/taskcluster/aws-provisioner/blob/5a2bc7c57b20df00f9c4357e0daeb7967e6f5ee8/lib/worker-type.js#L607-L624
+type AWSProvisionerUserData struct {
+	Data                BootstrapConfig `json:"data"`
+	Capacity            int             `json:"capacity"`
+	WorkerType          string          `json:"workerType"`
+	ProvisionerID       string          `json:"provisionerId"`
+	Region              string          `json:"region"`
+	AvailabilityZone    string          `json:"availabilityZone"`
+	InstanceType        string          `json:"instanceType"`
+	SpotBid             float64         `json:"spotBid"`
+	Price               float64         `json:"price"`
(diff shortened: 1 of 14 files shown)


workermanager.go
```diff
@@ -0,0 +1,85 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Could not register worker: %v", err)
+	}
+
+	c.AccessToken = reg.Credentials.AccessToken
+	c.Certificate = reg.Credentials.Certificate
+	c.ClientID = reg.Credentials.ClientID
+
+	// TODO: process reg.Expires
```
imbstack: Should we do this todo before this lands?

workermanager.go
```diff
@@ -0,0 +1,85 @@
...
+		if err != nil {
+			return "", fmt.Errorf("**** Can't reach worker-manager to see if there is a new deploymentId: %v", err)
+		}
+		workerManagerConfig := new(WorkerManagerConfig)
+		err = json.Unmarshal(wpfd.Config, &workerManagerConfig)
+		if err != nil {
+			return "", errors.New("WARNING: Can't decode /userData portion of worker type definition - probably somebody has botched a worker type update - not shutting down as in such a case, that would kill entire pool!")
+		}
+
+		publicHostSetup, err := workerManagerConfig.WorkerConfig.PublicHostSetup()
```
imbstack: Is this actually going to work with w-m? I don't see how the config as it is would work with this.

awsprovisioner.go
```diff
@@ -0,0 +1,85 @@
+package main
```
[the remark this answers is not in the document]
petemoore: The code got shuffled around, so this is just the stuff that pertains to the AWS Provisioner.

fileutil/file.go
```diff
@@ -0,0 +1,117 @@
+package fileutil
```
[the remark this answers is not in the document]
petemoore: Yes.

workermanager.go
petemoore: I don't see this as critical. Workers don't live too long, and if it used bad creds, it would get an error and shut itself down, and if it was running a task, resolve it as internal-error, and the task would get rescheduled for another worker, so it isn't critical, and also should only happen rarely for very long-lived workers. Workers typically shut down after 15 mins if they don't have a job, so quite unlikely.

workermanager.go
petemoore: Yes, it is tested via the CI of this PR itself, which is deployed with the code from this branch. It works the same way it worked with AWS Provisioner, the worker config contains a section for the generic-worker config file, but also supports providing other data for bootstrapping the worker.

Issue #244, opened 2019-10-09
Transition `.taskcluster.yml` from v0 to v1


owlishDeveloper:
Documentation on v0: [click](https://taskcluster-ui.herokuapp.com/docs/reference/integrations/github/taskcluster-yml-v0)
Documentation for v1: [click](https://taskcluster-ui.herokuapp.com/docs/reference/integrations/github/taskcluster-yml-v1)

owlishDeveloper:
@clrmachado This issue is yours!

owlishDeveloper:
An example https://github.com/taskcluster/json-parameterization/pull/9/files

clrmachado:
Thanks @owlishDeveloper!

clrmachado:
I've noticed that there is no replacement for `event.head.user.email ` any particular user I should set it to or should I use `nobody@mozilla.com` like in the shared example?

PR #248, opened 2019-10-19
Implement #244 - Update taskcluter.yml to v1

clrmachado:
@owlishDeveloper, so you can have a look at the yml file I've been working on! 
Thanks!


.taskcluster.yml
```diff
@@ -1,40 +1,47 @@
...
+version: 1
+policy:
+  pullRequest: public
 # Top level metadata is always required
-metadata:
-  name: "Taskcluster Generic Worker Tests"
-  description: "These tests should ensure that any new commits against generic worker codebase are tested across all supported worker types in gekco."
-  owner: "{{ event.head.user.email }}" # the user who sent the pr/push e-mail will be inserted here
-  source: "{{ event.head.repo.url }}"  # the repo where the pr came from will be inserted here
+#metadata:
```
owlishDeveloper: top-level metadata is not needed in v1, so this can go. The comments can be left tho, like this:

```
##############
# Taskcluster Generic Worker Tests
#
# These tests should ensure that any new commits against generic worker codebase are tested across all supported worker types in gecko.
# 
##############
```

.taskcluster.yml
```diff
@@ -1,40 +1,47 @@
...
-          - pull_request.synchronize
-          - push
-    scopes:
-      - generic-worker:cache:generic-worker-checkout
-      - secrets:get:repo:github.com/taskcluster/generic-worker
-    payload:
-      features:
-        taskclusterProxy: true
-      maxRunTime: 3600
+  $let:
```
owlishDeveloper: 👍

.taskcluster.yml
```diff
@@ -1,40 +1,47 @@
...
+    owner: "taskcluster-internal@mozilla.com" # the user who sent the pr/push e-mail will be inserted here
+    repo:
+      $if: 'tasks_for == "github-pull_request" && event["action"] in ["opened", "synchronize"]'
+      then:
+        url: ${event.pull_request.head.repo.clone_url}
```
owlishDeveloper: Instead of having variables in an object structure (`repo.url`), we can just declare variables with names like `repo_url` or something similar, or even just `repository`. This might take care of `InterpreterError at template.tasks.ref: object has no property ref`

.taskcluster.yml
```diff
@@ -1,40 +1,47 @@
...
+        ref: ${event.ref}
+        after: ${event.after}
+
+  in:
+    ##########################################################
+    ######### Windows Server 2012 R2 Multiuser Build #########
+    ##########################################################
+    - provisionerId: pmoore-test
+      workerType: win2012r2-cu
+      created: {$fromNow: ""}
```
owlishDeveloper: `created` is set by Taskcluster, so we don't need to set it here

.taskcluster.yml
```diff
@@ -1,40 +1,47 @@
...
+        description: This builds and tests the 64 bit Windows version of generic-worker (multiuser engine) on Windows Server 2012 R2
+        owner: ${owner} #the user who sent the pr/push e-mail will be inserted here
+        source: ${repo.url}  # the repo where the pr came from will be inserted here
+      scopes:
+        - generic-worker:cache:generic-worker-checkout
+        - secrets:get:repo:github.com/taskcluster/generic-worker
+      payload:
+        features:
+          taskclusterProxy: true
+        maxRunTime: 3600
```
owlishDeveloper: It seems to me `command` section (starting on line 45) has the same indentation as `payload` section (starts on line 41)? It should have the same indentation as lines 42 and 44. Same for `artifacts` and `mounts`. All similar spots in the files need to be checked as well

clrmachado: indeed, the indentation was completely off! I've updated it, but keep on getting the same error when I try running the validator tool

petemoore:
Many thanks @clrmachado for the PR, and apologies I didn't get to it sooner.

Also huge thanks to @owlishDeveloper for helping out with the review. Very much appreciated! 👍

petemoore:
@clrmachado I believe the merge conflict came from 297253f71464b136f5ff8a4595e31cd5e7ba441a so I resolved with commit 060e8d60fce57303a7bb5fac181449ee17ade9c0 (which also accidentally removed the last new line character of the file, but that shouldn't do any harm).

petemoore:
We've now upgraded to v1 in #246. Many thanks @clrmachado for the work you did here, I'm sorry it didn't make it in in the end, but your effort is still very much appreciated. Thank you.

PR #252, opened 2019-10-29
Bug 1592233 - calculate UID for new user under macOS multiuser

petemoore:
Whoops!

See [bug 1592233](https://bugzil.la/1592233) for context. These two lines [used to exist](https://github.com/taskcluster/generic-worker/commit/814aa9e8afee3e925b1295eeda12841da8754404#diff-db028d655af10079249eb4d4a2b5120cL29-L30) but got deleted because I thought they were no longer needed, but it seems they were.

Since I hadn't upgraded any mac workers, I hadn't spotted that this was broken, and unfortunately the unit and integration tests didn't catch it.

### DIFF
--- a/runtime/runtime_darwin.go
+++ b/runtime/runtime_darwin.go
@@ -26,6 +26,8 @@ func (user *OSUser) CreateNew(okIfExists bool) (err error) {
 		fullname="${0}"
 		password="${1}"
 		echo "Creating user '${username}' with home directory '${homedir}' and password '${password}'..."
+		maxid=$(dscl . -list '/Users' 'UniqueID' | awk '{print $2}' | sort -ug | tail -1)
+		uid=$((maxid+1))
 		/usr/bin/sudo dscl . -create "/Users/${username}"
 		/usr/bin/sudo dscl . -create "/Users/${username}" 'UserShell' '/bin/bash'
 		/usr/bin/sudo dscl . -create "/Users/${username}" 'RealName' "${fullname}"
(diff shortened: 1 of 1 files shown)


runtime/runtime_darwin.go
```diff
@@ -26,6 +26,8 @@ func (user *OSUser) CreateNew(okIfExists bool) (err error) {
 		fullname="${0}"
 		password="${1}"
 		echo "Creating user '${username}' with home directory '${homedir}' and password '${password}'..."
+		maxid=$(dscl . -list '/Users' 'UniqueID' | awk '{print $2}' | sort -ug | tail -1)
```
djmitche: does `sort` need `-n`?

milescrabill: looks like `sort -n` and `sort -g` are [roughly equivalent](https://askubuntu.com/questions/840781/sort-command-g-versus-n-flag) but `-n` is more common, so maybe more clear

petemoore: That should be covered by the `-g`:
```
     -g, --general-numeric-sort, --sort=general-numeric
             Sort by general numerical value.  As opposed to -n, this option handles general floating points.  It has a more permissive format than that allowed by -n but it has a significant per-
             formance drawback.
```

Perhaps `-n` would have been enough, but `-g` also covers it.

petemoore: Looks like pretty similar performance:

```
pmoore@Petes-iMac:~/go/src/github.com/taskcluster/generic-worker/worker_types master $ time for ((x=0; x<100;x++)); do dscl . -list '/Users' 'UniqueID' | awk '{print $2}' | sort -ug >/dev/null; done

real	0m3.442s
user	0m2.385s
sys	0m1.676s
pmoore@Petes-iMac:~/go/src/github.com/taskcluster/generic-worker/worker_types master $ time for ((x=0; x<100;x++)); do dscl . -list '/Users' 'UniqueID' | awk '{print $2}' | sort -un >/dev/null; done

real	0m3.405s
user	0m2.349s
sys	0m1.610s
```

PR #256, opened 2019-11-05
Allow specifying a default config.

tomprince:


petemoore:
I think the problem that you are looking to solve here is that there are certain generic-worker configuration settings which are specific to a machine image, for example the location of taskcluster-proxy executable on the machine, that you would prefer not to include in the worker pool definition, but rather have a config file burned into the machine image, and the worker pool definition only contain supplementary config which is not specific to the machine image, but specific to the worker pool.

My experience has been, that as soon as worker config is potentially located in several sources, it becomes much more difficult for people to understand why a worker has a particular config setting. Most users will not have access to server logs, so will not have a complete overview of the worker config. By placing some of it in the machine image, it effectively becomes opaque to most users, and troubleshooting issues can be very difficult, because they cannot determine what all the worker config settings are.

In contrast, if the config comes from a single source (namely the worker pool definition) the worker environment is immediately transparent. Every config setting can be seen, and reasoned about, for every user, since the config is public information.

Since worker config is managed by ci-admin in the firefox ci cluster, my proposal would be that you include a data structure to store information about machine image sets (perhaps named `imagesets.yml`), with content like this:


```yaml
docker-worker:
  worker-implementation: docker-worker
  gcp:
    image: projects/taskcluster-imaging/global/images/docker-worker-gcp-googlecompute-2019-11-04t22-31-35z

generic-worker-win2012r2:
  worker-implementation: generic-worker
  aws:
    amis:
      us-east-1: ami-04ff4e4c220abce54
      us-west-1: ami-070ee00d395f493d3
      us-west-2: ami-02161407768d981ea
  gcp:
    image: projects/pmoore-dev/global/images/win2012r2-rzuhq8vonqvbsgvhdqsd
  workerConfig:
    generic-worker:
      config:
        ed25519SigningKeyLocation: C:\generic-worker\generic-worker-ed25519-signing-key.key
        livelogExecutable: C:\generic-worker\livelog.exe
        taskclusterProxyExecutable: C:\generic-worker\taskcluster-proxy.exe
        workerTypeMetadata:
          machine-setup:
            maintainer: pmoore@mozilla.com
            script: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/win2012r2/bootstrap.ps1

deepspeech-win2012r2:
  worker-implementation: generic-worker
  aws:
    us-east-1: ami-00ac6736f325c15b8
    us-west-1: ami-000f332719212504f
    us-west-2: ami-032a6d4f0021ad2a9
  workerConfig:
    generic-worker:
      config:
        ed25519SigningKeyLocation: C:\generic-worker\generic-worker-ed25519-signing-key.key
        livelogExecutable: C:\generic-worker\livelog.exe
        taskclusterProxyExecutable: C:\generic-worker\taskcluster-proxy.exe
        workerTypeMetadata:
          machine-setup:
            maintainer: alissy@mozilla.com
            script: https://raw.githubusercontent.com/taskcluster/generic-worker/2509a1f8589b0b536186f5fe8f043385bf67197f/worker_types/deepspeech-win/bootstrap.ps1

generic-worker-linux:
  worker-implementation: generic-worker
  aws:
    us-east-1: ami-000f78733862c228b
    us-west-1: ami-0d2bc4a97a0aaa9d3
    us-west-2: ami-00e7eeb8d7449b1ce
  workerConfig:
    generic-worker:
      config:
        ed25519SigningKeyLocation: /etc/generic-worker/ed25519_key
        livelogExecutable: /usr/local/bin/livelog
        taskclusterProxyExecutable: /usr/local/bin/taskcluster-proxy
        tasksDir: /home
        workerTypeMetadata:
          machine-setup:
            maintainer: pmoore@mozilla.com
            script: https://raw.githubusercontent.com/taskcluster/generic-worker/b61e0d09401b0f4c4a8f33a5e2eb239bb90dc7ac/worker_types/gwci-linux/bootstrap.sh
```

When new machine images are created, it should be relatively straightforward to maintain information about the default config settings you want, and this also makes the information _explicit_ and _public_.

Once you have a data structure like this, it is relatively easy to associate a worker pool to an image set, and build the complete worker pool definition including _all_ the worker config settings.

This way you provide insight to people maintaining worker pool definitions, to know what the default config settings are for given machine images, and you make it transparently available to users with ease, making it much simpler for them to reason about the behaviour of a worker pool. It also makes it simple to have different sets of defaults for the same machine images, for example if you have a single machine image that has multiple versions of livelog / taskcluster-proxy installed, or has multiple signing keys burned in, so that the same image can be shared across worker pools that have different signing keys etc.

djmitche:
Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1592844

petemoore:
Note, I recently did a cleanup of config in the community cluster, and realised there is only a single config setting with a file path that is image-specific that doesn't have a reasonable default: [`ed25519SigningKeyLocation`](https://github.com/taskcluster/generic-worker/blob/v16.5.5/usage.go#L91). All other config settings can easily be avoided by just installing in standard locations. In the community cluster, we have significantly reduced the amount of config by doing this, and I plan to make a OS-specific default for `ed25519SigningKeyLocation` too in a release soon. Then there is no need for any image-specific config in the worker pool definitions.

See the [community cluster imagesets](https://github.com/mozilla/community-tc-config/blob/master/config/imagesets.yml) config file as an example.

petemoore:
> Note, I recently did a cleanup of config in the community cluster, and realised there is only a single config setting with a file path that is image-specific that doesn't have a reasonable default: [`ed25519SigningKeyLocation`](https://github.com/taskcluster/generic-worker/blob/v16.5.5/usage.go#L91). All other config settings can easily be avoided by just installing in standard locations. In the community cluster, we have significantly reduced the amount of config by doing this, and I plan to make a OS-specific default for `ed25519SigningKeyLocation` too in a release soon. Then there is no need for any image-specific config in the worker pool definitions.
> 
> See the [community cluster imagesets](https://github.com/mozilla/community-tc-config/blob/master/config/imagesets.yml) config file as an example.

@tomprince

PR #257, opened 2019-11-05
Bug 1593483 part 2

petemoore:
See [bug 1593483](https://bugzil.la/1593483) for additional context.

I think this should do it. It is pretty tricky to test.

The problem we had before was that if config file was found on the file system, we avoided the config negotiation with AWSProvisioner/WorkerManager and used the config file we had persisted. However, the function to check for a new `deploymentId` was therefore not updated, so the check for a new `deploymentId` was using the default function which just checked the local generic-worker config file, rather than talking to AWSProvisioner/WorkerManager.

The way I've solved this turned out to be a bit of a rewriting, so that we have a `gwconfig.Provider` interface that has a method for updating generic-worker config, and for fetching the latest deployment ID.

### DIFF
--- a/gwconfig/provider.go
+++ b/gwconfig/provider.go
@@ -1,0 +1,14 @@
+package gwconfig
+
+// gwconfig.Provider is an interface for configuring generic-worker.
+// Implementors include AWSProvisioner, AWSProvider, GCPConfigProvider, and
+// gwconfig.File for configuring generic-worker from a local file.
+type Provider interface {
+	// NewestDeploymentID fetches the latest/newest/most-recent deployment ID.
+	// Depending on the provider, this may make a network request to retrieve
+	// data from a taskcluster entity (such as a worker pool definition) or may
+	// just look at the local generic-worker config file. See generic-worker
+	// help text for information about the deploymentId property.
+	NewestDeploymentID() (string, error)
+	UpdateConfig(c *Config) error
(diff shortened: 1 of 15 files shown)


gwconfig/provider.go
```diff
@@ -0,0 +1,14 @@
+package gwconfig
+
+// gwconfig.Provider is an interface for configuring generic-worker.
+// Implementors include AWSProvisioner, AWSProvider, GCPConfigProvider, and
+// gwconfig.File for configuring generic-worker from a local file.
+type Provider interface {
```
djmitche: Maybe this should be ConfigProvider or ConfigSource to avoid confusion with worker-manager or worker-runner providers?

petemoore: Funnily enough in my first draft I called it ConfigProvider, but then I changed it based on [this advice](https://blog.golang.org/package-names):

> __Avoid stutter__. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity. 

Anything using Provider refers to it as `gwconfig.Provider` so hopefully it should be clear that it is a config provider. But I also wasn't totally happy with "Provider" either as it sounds like a Worker Manager term.

PR #262, opened 2019-12-13
Added --configure-for-azure option for running in Azure cloud

petemoore:
This PR adds `--configure-for-azure` option to generic-worker to unblock Azure work.

### DIFF
--- a/azure.go
+++ b/azure.go
@@ -1,0 +1,169 @@
+package main
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"net/http"
+
+	"github.com/taskcluster/generic-worker/gwconfig"
+	"github.com/taskcluster/httpbackoff"
+)
+
+var (
+	// not a const, because in testing we swap this out
+	AzureMetadataBaseURL = "http://169.254.169.254"
+)
+
+type (
+	AzureConfigProvider struct {
+	}
+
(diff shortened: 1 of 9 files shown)


azure.go
```diff
@@ -0,0 +1,169 @@
...
+	err = json.Unmarshal(attestedDoc, &doc)
+	if err != nil {
+		return fmt.Errorf("Could not unmarshal attested document %q into AttestedDocument struct - is it valid JSON? %v", string(attestedDoc), err)
+	}
+	if doc.Encoding != "pkcs7" {
+		return fmt.Errorf(`Attested document has unsupported encoding (%q) - generic-worker only supports "pkcs7"`, doc.Encoding)
+	}
+
+	// TODO: when Azure Provider is completed in Worker Manager, we'll be able
+	// to do the following, but that isn't available yet:
```
imbstack: This is just because the schema isn't there yet to generate this from?

petemoore: That's right.

djmitche:
This is ridiculous.  Let's talk as a team, or just you and me and coop, and settle this.

djmitche:
I scheduled a meeting on Monday.

petemoore:
It is likely we won't need to use this option, and we can use taskcluster-worker-runner instead. But rather than leave this PR open, or delete the code which may be useful if we hit issues with taskcluster-worker-runner, I'll merge for now. We can always remove this code later if it isn't needed, but no point in deleting it before we have a working alternative.

PR #264, opened 2019-12-17
Bug 1604201 - allow redirecting output to a named pipe

djmitche:


main.go
```diff
@@ -121,6 +121,19 @@ func main() {
...
+		// include worker-runner protocol traffic, but for the moment it simply
+		// provides a way to channel generic-worker logging to worker-runner
+		if arguments["--worker-runner-protocol-pipe"].(string) != "" {
+			protocolPipe := arguments["--worker-runner-protocol-pipe"].(string)
+			f, err := os.OpenFile(protocolPipe, os.O_RDWR, 0)
+			exitOnError(CANT_CONNECT_PROTOCOL_PIPE, err, "Cannot connect to %s: %s", protocolPipe, err)
+
+			os.Stdin = f
+			os.Stdout = f
+			os.Stderr = f
```
petemoore: Interesting, so the pipe can read from and written to with regular `os.File` semantics? I guess the returned file isn't seekable, but is readable and writable, and that reads are read from the pipe, and writes are written to the pipe? Are parallel writes from multiple goroutines atomically sequenced?

djmitche: Yeah, it works basically the same as stdio streams -- individual calls to `write` are sequenced properly.

PR #266, opened 2020-01-09
Bug 1607606 - run the worker-runner protocol on stdio

djmitche:


main.go
```diff
@@ -225,6 +235,16 @@ func main() {
 	}
 }
 
+func initializeProtocol(stdin io.Reader, stdout io.Writer) {
```
petemoore: Nit: can we rename `stdin` -> `in` and `stdout` -> `out` since they don't need to be standard in and standard out, but any readers/writers?

main.go
```diff
@@ -225,6 +235,16 @@ func main() {
...
+	workerRunnerTranport := protocol.NewStdioTransport()
```
petemoore:
```suggestion
	workerRunnerTransport = protocol.NewStdioTransport()
```

main.go
```diff
@@ -225,6 +235,16 @@ func main() {
...
+
+	go io.Copy(workerRunnerTranport, stdin)
```
petemoore:
```suggestion
	go io.Copy(workerRunnerTransport, stdin)
```

main.go
```diff
@@ -225,6 +235,16 @@ func main() {
...
+	go io.Copy(stdout, workerRunnerTranport)
```
petemoore:
```suggestion
	go io.Copy(stdout, workerRunnerTransport)
```

main.go
```diff
@@ -225,6 +235,16 @@ func main() {
...
+
+	WorkerRunnerProtocol = protocol.NewProtocol(workerRunnerTranport)
```
petemoore:
```suggestion
	WorkerRunnerProtocol = protocol.NewProtocol(workerRunnerTransport)
```

main.go
```diff
@@ -136,6 +144,8 @@ func main() {
 			os.Stderr = f
 		}
 
+		initializeProtocol(os.Stdin, os.Stdout)
```
petemoore: Nit: can we call this `initializeWorkerRunnerProtocol`?

main.go
djmitche: !!! that is a bug I wish go would warn me about!

main_test.go
```diff
@@ -252,3 +254,42 @@ func TestUsage(t *testing.T) {
...
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+
+	// check that we sent a "hello" message in response
+	var msg map[string]interface{}
+	err := json.Unmarshal(writer.written[1:], &msg)
+	assert.NoError(t, err)
+	assert.Equal(t, "hello", msg["type"])
+}
```
petemoore: Nice!

